PR #9, opened 2023-09-21
feat: Add default timezone for parsers BNCH-74231

nathan-stender:
Some instruments do not add timezones to their timestamps.

The draft 2020-12 json schema format that ASM uses requires timestamps to be valid against the [rfc3339](https://datatracker.ietf.org/doc/html/rfc3339) standard, which requires a timezone.

To fix this, we automatically add a default timezone any timestamp that does not have one already. 

This default timezone can be specified by the user when creating a parser via the `ParserFactory`, and defaults to `UTC`

Note that as part of this change, `validate_schema` in `test_utils` now runs against `format_checker=jsonschema.validators.Draft202012Validator.FORMAT_CHECKER`, which previously caused the tests to fail due to invalid timestamps, now all pass.


README.md
```diff
@@ -27,7 +27,7 @@ Convert a file to an ASM dictionary:
 from allotropy.parser_factory import VendorType
 from allotropy.to_allotrope import allotrope_from_file
 
-asm_schema = allotrope_from_file("filepath.txt", VendorType.MOLDEV_SOFTMAX_PRO)
+asm_schema = allotrope_from_file("filepath.txt", Vendor.MOLDEV_SOFTMAX_PRO)
```
nathan-stender, line 30: Note to reviewer: this change is fixing a typo I noticed

src/allotropy/allotrope/schemas/cell-culture-analyzer/BENCHLING/2023/09/cell-culture-analyzer.json
```diff
@@ -17,7 +17,7 @@
                     "$asm.property-class": "http://purl.allotrope.org/ontologies/result#AFR_0002095",
                     "$asm.pattern": "value datum",
                     "$asm.type": "http://www.w3.org/2001/XMLSchema#dateTime",
-                    "$ref": "#/$defs/tStringValue"
+                    "$ref": "#/$defs/tDateTimeValue"
```
nathan-stender, line 20: Note to reviewer: this change is fixing a mistake in the schema I found.

src/allotropy/parsers/utils/timestamp_parser.py
```diff
@@ -0,0 +1,37 @@
+from datetime import timedelta, timezone
+from typing import Optional
+
+from dateutil import parser
+import pytz
+
+from allotropy.allotrope.allotrope import AllotropeConversionError
+
+TIMEZONE_CODES_MAP = {
```
stephenworlow, line 9: Do you think we'll need Europe?

stephenworlow (approved):

PR #20, opened 2023-09-28
feat: redefine calculated document building for relative standard curve in AppBio Quantstudio

alejandro-salgado:
- Decouple calculated document builder in AppBio Quantstudio
- Redefine way to construct calculated documents for relative standard curve in AppBio Quantstudio

### DIFF
--- a/src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_builders.py
+++ b/src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_builders.py
@@ -10,7 +10,8 @@ from allotropy.allotrope.models.pcr_benchling_2023_09_qpcr import (
     ExperimentType,
 )
 from allotropy.parsers.appbio_quantstudio.appbio_quantstudio_calculated_documents import (
-    CalculatedDocumentBuilder,
+    build_quantity,
+    iter_calculated_data_documents,
 )
 from allotropy.parsers.appbio_quantstudio.appbio_quantstudio_structure import (
     AmplificationData,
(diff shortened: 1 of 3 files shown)


bdworth (changes requested):
Initial high-level questions before a more careful review: 

1. If this PR is part refactor and part feature changes as the description suggests, would it be possible to split out the refactor(s) into separate PR(s)?
2. Are there new tests in the works for the new functionality?

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_builders.py
```diff
@@ -623,25 +624,23 @@ def build(reader: LinesReader) -> Data:
                     )
                 )
 
-            well.calculated_document = CalculatedDocumentBuilder.build_quantity(
-                well_item
-            )
-
-        endogenous_control = get_str(results_metadata, "Endogenous Control")
-        reference_sample = get_str(results_metadata, "Reference Sample")
+            well.calculated_document = build_quantity(well_item)
```
bdworth, line 627: pre-existing bug that should nonetheless be fixed: if `well.items.values()` is empty, `well_item` have been assigned.

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_builders.py
```diff
@@ -623,25 +624,23 @@ def build(reader: LinesReader) -> Data:
...
 
-        calc_doc_builder = CalculatedDocumentBuilder(
-            wells=wells,
-            experiment_type=header.experiment_type,
-            r_target=endogenous_control or "",
-            r_sample=reference_sample or "",
-        )
+        endogenous_control = get_str(results_metadata, "Endogenous Control") or ""
+        reference_sample = get_str(results_metadata, "Reference Sample") or ""
```
bdworth, line 630: Please don't include refactors with functional changes. I.e., leave these lines as is and only change the `calc_doc_builder` call.

Out of scope: `get_str` has a `default` param, but it's not very useful for type checking since it can be optional and the function returns `Optional[str]`. If we had a utility method like `get_str_or_default` where `default` was required and `str` was returned, we could use that here.

bdworth (changes requested):
Sorry, but these changes are hard to follow. It seems like a lot is just refactoring. Please split the refactors into their own PR(s).

PR #35, opened 2023-10-09
test: add broken calculated document test for AppBio Quantstudio

alejandro-salgado:
- Add test for broken calculated document structure in AppBio QuantStudio


tests/parsers/appbio_quantstudio/appbio_quantstudio_parser_test.py
```diff
@@ -60,6 +62,11 @@ def test_parse_appbio_quantstudio_to_asm_contents(output_file: str) -> None:
             get_rel_std_curve_data(),
             get_rel_std_curve_model(),
         ),
+        (
```
nathan-stender, line 65: Can you add a comment explaining what broken means here (I assume here broken means a broken chain of references - I think broken is a fine name, but it won't be clear to someone coming in to this for the first time, so it would be good to leave a few sentence comment explaining what this test is doing)

alejandro-salgado: I added a comment explaining the objective of the test

PR #37, opened 2023-10-12
refactor: Replace datetime.timezone with ZoneInfo

nathan-stender:
ZoneInfo more easily accepts named timezones, so this change will make creating and passing in a default_timezone easier.

Also adds a test confirming daylight savings time adjustment works as expected.


mdlu (approved):
there's a mypy error that needs to be fixed because of a recent mypy version upgrade, but I'm going to hold off on fixing it until I learn how the permissions on this repo work

`src/allotropy/allotrope/schema_parser/generate_schemas.py:7: error: Unused "type: ignore" comment, use narrower [import-untyped] instead of [import] code  [unused-ignore]`

mdlu (changes requested):
see above

nathan-stender:
@mdlu it seems re-running the check fixed the issue :D

PR #38, opened 2023-10-12
refac: replace builder classes with create methods in AppBio QuantStudio

alejandro-salgado:
- Replace builder classes with create methods in AppBio QuantStudio structure where possible
- Add utils to get data from pandas series


src/allotropy/parsers/utils/values.py
```diff
@@ -13,6 +15,14 @@ def try_int(value: Optional[str]) -> Optional[int]:
         return None
 
 
+def assert_int(value: Optional[str], name: Optional[str] = None) -> int:
+    return assert_not_none(
+        try_int(value),
+        name,
+        msg="Expected int value {f'for {name}' if name else ''}",
```
nathan-stender, line 22:
```suggestion
        msg=f"Expected int value{f' for {name}' if name else ''}",
```

nathan-stender (changes requested):
One quick change - this error string formatting won't work as is

nathan-stender (approved):

src/allotropy/parsers/utils/values.py
```diff
@@ -42,3 +52,66 @@ def assert_not_none(
...
+
+
+def assert_not_empty_df(df: pd.DataFrame, msg: str) -> pd.DataFrame:
+    if df.empty:
+        raise AllotropeConversionError(msg)
+    return df
+
+
+def str_from_series(
+    data: pd.Series, key: str, default: Optional[str] = None
```
bdworth, line 71: The `default` parameter is unused in callers.

And if we did want a default, it would be better to have a different signature (similar to `assert_str_from_series`) in which default wasn't `Optional` and thus neither was the return type.

bdworth (changes requested):

bdworth (changes requested):
More generally, this is a very large PR and thus hard to review. Could you please split up? At least into the two components called out in the current description?
```
(1) Replace builder classes with create methods in AppBio QuantStudio structure where possible
(2) Add utils to get data from pandas series
```

alejandro-salgado:
Close current PR to create requested smaller PRs

PR #44, opened 2023-10-16
Rename 'fix' to 'fmt' to match Hatch default

bdworth:
"Fix" may be more self-evidently named, but better to stick with Hatch default command keyword, especially given that all other `lint` commands use the default keywords.

Note that CI checks run `hatch run lint:all`, which does not run `fmt` but `style`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ style = [
   "black --check --diff {args:.}",
   "flake8 .",
 ]
-fix = [
+fmt = [
   "black {args:.}",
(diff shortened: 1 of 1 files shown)


nathan-stender (changes requested):
Can we add a new command like `autofix` that does `fmt` and `typing`? 

I'm ok with going back to fmt, but I'd prefer to still have a more intuitive cmd.

bdworth:
But `typing` doesn't do any auto-fixing 😕

nathan-stender (approved):

PR #63, opened 2023-10-23
Added AllotropyException to be used instead of Generic Exception

EihabSyed:


src/allotropy/allotrope/allotrope.py
```diff
@@ -19,6 +19,10 @@ class AllotropeConversionError(Exception):
     pass
 
 
+class AllotropeException(Exception):
```
mdlu, line 22: could you rename this `AllotropyException`?

mdlu (changes requested):
this is a great start, could you also replace all instances of `Exception` in the repo, not just in the Envision parser?

slopez-b (commented):
Any reason to create a new exception instead of using the existing `AllotropeConversionError`?

EihabSyed:
> Any reason to create a new exception instead of using the existing `AllotropeConversionError`?

The `AllotropeConversionError` was a specific error for conversion whereas the new `AllotropyException` is meant to provide more info as to where the error originates from, in this case the the error is from `Allotropy`

mdlu (dismissed):

mdlu (approved):

PR #82, opened 2023-10-24
Yukthiw/parse checksum example weyland yutani parser

Yukthiw:


gvwilson:
happy to take this one on

PR #83, opened 2023-10-24
feat: Update PerkinElmer EnVision parser to match new plate reader schema

mdlu:
This updates the EnVision adapter to match the newest plate reader schema, merging absorbance/fluorescence/luminescence into the same schema and supporting all modes. This includes some other smaller changes to support this:

- `undefined sample role` is removed from the list of SampleRoleTypes, as it is no longer a valid option in the schema
- new test files from Deep Genomics are now supported in testing
- `PE_Envision_fluorescence_example01_v2.json` is deleted, as it was being used as an example to match to; `PE_Envision_fluorescence_example01.json` has now been updated to match that example

Product requirements (Benchling-internal): https://docs.google.com/document/d/1WfmQ-4H9m1MlsZ5uevAwwRmrGh-Pv9rjEMP6jf3qCQY/edit#heading=h.opfiwvlejdqk


src/allotropy/parsers/utils/values.py
```diff
@@ -20,6 +20,15 @@ def try_float(value: Optional[str]) -> Optional[float]:
         return None
 
 
+def try_str(value: Any) -> Optional[str]:
+    if not value:
+        return None
+    try:
+        return str(value or "")
+    except ValueError:
+        return None
```
bdworth, line 29: How could `str` of any non-`None` value throw `ValueError`? Also, since `value` is guaranteed to be truthy here, `or ""` will never get hit.
```suggestion
    return str(value)
```

Also, could you please write tests for this utility method? Integration tests are nice but shouldn't contribute to coverage.

tests/parsers/perkin_elmer_envision/perkin_elmer_envision_parser_test.py
```diff
@@ -36,7 +39,22 @@ def test_get_model() -> None:
     parser = PerkinElmerEnvisionParser(TimestampParser())
     model = parser._get_model(get_data())
 
-    if model.measurement_aggregate_document:
-        model.measurement_aggregate_document.measurement_identifier = ""
+    # remove all random UUIDs
+    if model.plate_reader_aggregate_document:
```
bdworth, line 43: Could you use the walrus operator here to improve readability?

bdworth (dismissed):
I didn't review everything here but have two quick suggestions.

LimulusPolyphemus:
@mdlu I noticed a minor issue with the fields in the `data system document`.

The `model number` should be "Envision", and the "Serial Number" which is currently being written to `model number` should be returned to `equipment serial number`

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py
```diff
@@ -169,16 +169,18 @@ def create(reader: CsvReader) -> BasicAssayInfo:
...
 
     @staticmethod
     def create(reader: CsvReader) -> PlateType:
         reader.drop_until("^Plate type")
         data = reader.pop_csv_block_as_df("^Plate type")
         data_df = assert_not_none(data, "Plate type").T
         return PlateType(
-            try_float_or_none(
-                str(df_to_series(data_df).get("Number of the wells in the plate"))
+            float(
```
nathan-stender, line 180: qq: why not `try_float` here?

nathan-stender (dismissed):

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py
mdlu: good question, `number_of_wells` is required in the ASM now I think, but as far as I can tell this line always exists?

nathan-stender (approved):

PR #85, opened 2023-10-25
Feat: add new rec cell counting schema

slopez-b:
Add cell-counting REC/2023/09 schema

The schema was renamed `BENCHLING/2023/11/cell-counting` since we have some library specific changes:

- Added `data system document` to the schema since it was not included in the allotrope version.
- Made `asset management identifier` optional since we don't have a value for it in the current cell counting adapters.


nathan-stender (approved):
Approved with small nit request

alejandro-salgado (approved):

PR #89, opened 2023-10-26
Governance model

gvwilson:
Please review (and let me know whether Nathan or Joe should be BDFL - I picked Nathan because he's been the largest contributor to date).


nathan-stender (approved):
I humbly accept this appointment for life 😈

PR #84, opened 2023-10-24
feat: add support for extracting processed data values and instrument spec

trevorHalumDG:
- Example parser now extracts the processed data values from the csv
- Fixed the layout of cell_data by transposing the data frame
- Added support for extracting the instrument spec

PR #97, opened 2023-10-27
feat: parse Weyland-Yutani files with optional checksums (refactoring CsvReader)

gvwilson:
This PR includes some changes made in #82 and #84 and should only be merged after that are.

1. Refactor `CsvReader` to allow caller to pass in a few optional arguments to e.g. specify what counts as an empty line (because a line containing nothing but commas and spaces is and empty row in the CSV sense, but not in the text sense).
2. Modify the Weyland-Yutani parser to use this to grab chunks as dataframes and extract data from them.
3. Add sample data to test that the Weyland-Yutani parser works on files that have an optional checksum field.

Note that the checksum is being added to the `BasicAssayInfo` portion of the result; please let me know if it would be more correct to put it in a calculated value field.


gvwilson:
happy to take this one on

PR #107, opened 2023-11-07
refac: restructure Agilent Gen5 with explicit dataclasses structure

alejandro-salgado:
- Refactor Agilent Gen5 with explicit dataclasses structure


src/allotropy/parsers/agilent_gen5/plate_data.py
```diff
@@ -358,12 +507,157 @@ def _blank_data_cube(self, measures: list[float]) -> TDatacube:
...
+
+    @staticmethod
+    def get_data_point_cls() -> type[DataPoint]:
+        raise NotImplementedError
+
     def to_allotrope(self, measurement_docs: list) -> Any:
         raise NotImplementedError
+
+
+class AbsorbancePlateData(PlateData):
```
nathan-stender, line 588: Can we drop these classes, and instead make build the model inside the parser? Instead of returning a class corresponding to read mode, just return the PlateData, and store the plate_type.read_mode to decide which model to build.


This has a couple purposes: 

1. It more closely follows the structure of the other parsers (structure classes hold data, parser builds the models)
2. It paves the way for the new plate reader schema, where all of these will use the same model.

nathan-stender (changes requested):

src/allotropy/parsers/lines_reader.py
```diff
@@ -9,11 +9,20 @@
 from allotropy.allotrope.allotrope import AllotropyError
 
 
+def _decode(bytes_content: bytes, encoding: Optional[str]) -> str:
```
nathan-stender, line 12: Is moving this out intentional or a merge issue?

nathan-stender (changes requested):

nathan-stender (approved):

nathan-stender (approved):

PR #105, opened 2023-11-06
up-to-date version of Weyland-Yutani parser example

gvwilson:
This PR replaces #80 - once it is in `main`, @gvwilson will merge other branches created at DG and start writing the tutorial.

cc @nathan-stender

PR #110, opened 2023-11-10
feat: add example parser for Weyland-Yutani data

gvwilson:
-   Add new parser type.
-   Add code to `lines_reader.py` to return dataframe.
-   Modify typing to make happy.
-   Add test data.

This supercedes a previous multi-commit effort in #105 that had become very tangled.


src/allotropy/parsers/lines_reader.py
```diff
@@ -36,74 +38,90 @@ def current_line_exists(self) -> bool:
     def get(self) -> Optional[str]:
         return self.lines[self.current_line] if self.current_line_exists() else None
 
-    def match(self, pattern: str) -> bool:
+    def match(self, match_pat: str) -> bool:
```
bdworth, line 41: If you're going to rename the parameters, could you spell out `pattern` rather than `pat`?

src/allotropy/parsers/utils/values.py
```diff
@@ -31,6 +33,15 @@ def natural_sort_key(key: str) -> list[str]:
 T = TypeVar("T")
 
 
+def assert_df(
```
bdworth, line 36: Couldn't this functionality be achieved using `assert_not_none`?

src/allotropy/parsers/lines_reader.py
```diff
@@ -8,6 +8,8 @@
 
 from allotropy.allotrope.allotrope import AllotropyError
 
+EMPTY_STR = r"^\s*$"
```
bdworth, line 11: `EMPTY_STR_PATTERN`? I'd expect an `EMPTY_STRING` constant to be equal to `""`.

src/allotropy/parsers/lines_reader.py
```diff
@@ -36,74 +38,90 @@ def current_line_exists(self) -> bool:
...
 
-    def match(self, pattern: str) -> bool:
+    def match(self, match_pat: str) -> bool:
         line = self.get()
-        return False if line is None else bool(search(pattern, line))
+        return False if line is None else bool(search(match_pat, line))
 
-    def is_empty(self) -> bool:
-        return self.match("^\\s*$")
+    def is_empty(self, empty_pat: str = EMPTY_STR) -> bool:
```
bdworth, line 45: I'm not sold on override the empty string pattern, but if we want to do it, I think  `empty_pattern` should be a field on `LinesReader` and not a parameter on various methods.

src/allotropy/parsers/example_weyland_yutani/example_weyland_yutani_parser.py
```diff
@@ -0,0 +1,89 @@
...
+from allotropy.parsers.vendor_parser import VendorParser
+
+
+class ExampleWeylandYutaniParser(VendorParser):
+    def _parse(self, raw_contents: IOBase, _: str) -> Model:
+        reader = CsvReader(raw_contents)
+        return self._get_model(Data.create(reader))
+
+    def _get_model(self, data: Data) -> Model:
+        if data.number_of_wells is None:
```
bdworth, line 36: maybe extract `data_number_of_wells` to a local var and use it here and below to be more explicit that it has been checked?

src/allotropy/parsers/example_weyland_yutani/example_weyland_yutani_structure.py
```diff
@@ -0,0 +1,101 @@
...
+
+from dataclasses import dataclass
+from typing import Optional
+
+import pandas as pd
+
+from allotropy.allotrope.allotrope import AllotropeConversionError
+from allotropy.parsers.lines_reader import CsvReader
+
+EMPTY_CSV_LINE = r"^[,\s]*$"
```
bdworth, line 11: I think this needs to be `^,*$`. We don't want lines like ",  ," to read as empty.

bdworth (changes requested):

src/allotropy/parsers/lines_reader.py
gvwilson: I chose `match_pat` to be consistent with e.g. `pop_if_match` and `drop_until` further down in the same file - `pattern` seemed to be inconsistent (?).

src/allotropy/parsers/utils/values.py
gvwilson: `assert_df` was created to make the type checker happy: without it, I get complaints in `src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py` because it isn't sure that the object in hand is not `None`, so it doesn't want me to call dataframe methods.

src/allotropy/parsers/lines_reader.py
gvwilson: thanks - fixed in the latest push

src/allotropy/parsers/lines_reader.py
gvwilson: I hadn't thought about making this a field of the object - I think having it as a parameter results in about the same number of changes to existing code, but am happy to refactor if you prefer the former.

src/allotropy/parsers/example_weyland_yutani/example_weyland_yutani_structure.py
gvwilson: Good catch - changed in the latest push.

mdlu (approved):

src/allotropy/parsers/utils/values.py
bdworth: ok, I'll take a look at this in follow-up.

bdworth (approved):

PR #118, opened 2023-11-10
Make try_int() raise rather than return None

bdworth:
Depends on #111


bdworth:
This PR is behaving oddly, so moved the first commit to #124.

PR #119, opened 2023-11-10
Rename try_float to try_float_or_none

bdworth:
Rename for similar reasons as `try_int` and `try_int_or_none` in 613b895e3e8794195c56acd629084179db76af06.

If we want to add a `try_float` later, it should fail rather than return `None`.


nathan-stender:
Is there some PEP standard or very common convention that motivates this change? 

I thought the name and use was pretty intuitive here. To me (and the implementer) try makes sense here as try to get a value and otherwise treat as optional (as we don't really need a special function to fail, other than to provide a more explicit error message)

bdworth:
This is a companion to #124, and I just added some elaboration there.

nathan-stender (approved):

PR #120, opened 2023-11-10
feat: adding scripts to go with Weyland-Yutani example

gvwilson:
Add scripts to generate Weyland-Yutani datafiles, convert the result of parsing them to dataframes, and convert a CSV to JSON.

cc @mdlu @nathan-stender


gvwilson:
@mdlu passes all checks - I think this one is ready. (I'm going to use these scripts in the tutorial.)

mdlu (approved):
:shipit:

PR #121, opened 2023-11-10
Remove assert_df

bdworth:
`assert_not_none` does the same thing ... albeit with different error message defaults, but that's not enough to justify keeping `assert_df`.


gvwilson:
Thanks - I introduced `assert_df` to get around some mypy complaints, but I was clearly on the wrong track. Very happy to see it yanked out.

nathan-stender (approved):

PR #132, opened 2023-11-16
refac: improve perkin elmer envision adapter with better type handling

alejandro-salgado:
- Improve Perkin Elmer Envision adapter with better type handling
- Improve Lines Reader with cleaner functionality

### DIFF
--- a/src/allotropy/parsers/lines_reader.py
+++ b/src/allotropy/parsers/lines_reader.py
@@ -53,9 +53,7 @@ class LinesReader:
         return line
 
     def pop_if_match(self, match_pat: str) -> Optional[str]:
-        if self.match(match_pat):
-            return self.pop()
-        return None
+        return self.pop() if self.match(match_pat) else None
 
     def pop_data(self) -> Optional[str]:
         self.drop_empty()
--- a/src/allotropy/parsers/lines_reader.py
+++ b/src/allotropy/parsers/lines_reader.py
@@ -103,11 +103,14 @@ class CsvReader(LinesReader):
         self, match_pat: Optional[str] = None, empty_pat: str = EMPTY_STR_PATTERN
     ) -> list:
         self.drop_empty(empty_pat)
+
+        if match_pat and not self.match(match_pat):
+            msg = f"Did not find {match_pat}"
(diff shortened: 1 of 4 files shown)


src/allotropy/parsers/lines_reader.py
```diff
@@ -53,28 +53,34 @@ def pop(self) -> Optional[str]:
...
 
     def pop_data(self) -> Optional[str]:
         self.drop_empty()
         return self.pop()
 
-    def drop_until(self, match_pat: str) -> Optional[str]:
+    def drop_until(
+        self,
+        match_pat: str,
+        inclusive: bool = False,  # noqa: FBT001, FBT002
```
nathan-stender, line 65: Actually don't want to ignore these QA checks, which generally guide use to avoid adding bool flags to functions. 

This is an anti-pattern to avoid because:

1. It can be hard to tell what a short-named bool flag does for future engineers
2. It can lead to functions with several bool flags changing the behavior of the function, making it hard to tell from the call what a function is doing.

The general good pattern here is to create two functions, one for each behavior, e.g. in this case you could make the function:

```
def drop_up_to(self, match_pat: str):
    self.drop_until(match_pat)
    self.pop()
```

The good pattern is to create two functions in this case, e.g. in this case you could split into

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py
```diff
@@ -54,34 +61,43 @@ class PlateInfo:
 
     @staticmethod
     def create(reader: CsvReader) -> Optional[PlateInfo]:
-        data = reader.pop_csv_block_as_df("^Plate information")
-        data_df = assert_not_none(data, "Plate information CSV block")
-        series = df_to_series(data_df).replace(np.nan, None)
+        assert_not_none(
+            reader.pop_if_match("^Plate information"),
+            "Expected to get plate information",
```
nathan-stender, line 66:
```suggestion
            "Expected 'Plate information' block"
```

nathan-stender (changes requested):
Some great cleanup here, thank you!

Couple changes requested

PR #141, opened 2023-11-27
Extract ContentsType

bdworth:
Extract a constant to better see places where `io.IOBase` is currently used for typing.

Out of scope: changing `ContentsType` to something else so we can remove type suppressions when passing vars of this type to `pd.read_csv`.


mdlu (changes requested):
looks like a couple of merge conflicts, will re-review when addressed

bdworth:
Made unnecessary by refactors culminating in #162

PR #144, opened 2023-11-28
Simplify vi_cell_blu_parser

bdworth:
We shouldn't be using `iterrows`. Not just because it's bad for perf (e.g., https://towardsdatascience.com/efficiently-iterating-over-rows-in-a-pandas-dataframe-7dd5f9992c01) but because it erases type information by returning a `Series` of values of different types.

This cleans up usage of `iterrows` in vi-cell-blu. Test util changes are needed to consider `int` and `np.int64` the same for tests since types are now preserved better in src.


src/allotropy/parsers/beckman_vi_cell_blu/vi_cell_blu_reader.py
```diff
@@ -1,70 +1,10 @@
...
-            if not isinstance(col, pd.Series):
-                continue
-
-            new_col = (
-                col if col.dtype == desired_type else conversors[desired_type](col)
-            )
-            columns.append(new_col)  # type: ignore [arg-type]
-
-        return pd.concat(columns, axis=1).replace(np.nan, None)
+        return cast(pd.DataFrame, pd.read_csv(contents, index_col=False))  # type: ignore[call-overload]
```
bdworth, line 10: cast is necessary because `read_csv` returns `DataFrame | TextFileReader`

mdlu (dismissed):
nice! just confirming that `desired_columns` isn't needed anymore because all the types can now be inferred?

bdworth:
Review became stale as soon as I changed this PR to target master. But I also split up the reader changes into #150.

mdlu (approved):
I still have the same question as above ^

bdworth:
@mdlu `desired_columns` is in the changes that I split out.

mdlu (approved):

PR #151, opened 2023-11-29
Add clearer user-facing error messaging

mdlu:
This PR clarifies error exception messages to be standardized and clear, so that they can be most helpful to be displayed to a user for troubleshooting purposes.

Also removes `AllotropyError`, simplifying all errors to be `AllotropeConversionError`.
(also adds EnVision updates to CHANGELOG, as it wasn't added previously)

Doc: https://benchling.quip.com/aQwUA54C85Me/User-Facing-Adapter-Errors


nathan-stender (dismissed):

src/allotropy/allotrope/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class AllotropeConversionError(Exception):
```
bdworth, line 1: thoughts on putting this in `src/allotropy/` so the import would be `from allotropy.exceptions`?

(regardless, it's a lot better than `from allotropy.allotrope.allotrope`!)

nathan-stender (approved):

bdworth (approved):
This is going to make downstream error handling and debugging much better. Thanks.

nathan-stender (approved):

PR #128, opened 2023-11-15
refac: improve way to extract and validate information from pandas series in AppBio QuantStudio

alejandro-salgado:
- Improve way to extract and validate information from pandas series in AppBio QuantStudio

* first part of https://github.com/Benchling-Open-Source/allotropy/pull/38

PR #152, opened 2023-11-29
refac: redefine builders as create methods in AppBio QuantStudio

alejandro-salgado:
- Refactor builders as create methods in AppBio QuantStudio

* first part at https://github.com/Benchling-Open-Source/allotropy/pull/128
* second part of https://github.com/Benchling-Open-Source/allotropy/pull/38


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -165,13 +469,200 @@ class Result:
...
+
 
 @dataclass
 class MeltCurveRawData:
     reading: list[float]
     fluorescence: list[Optional[float]]
     derivative: list[Optional[float]]
 
+    @staticmethod
+    def build(data: pd.DataFrame, well: Well) -> MeltCurveRawData:
```
nathan-stender, line 646:
```suggestion
    def create(data: pd.DataFrame, well: Well) -> MeltCurveRawData:
```

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_data_creator.py
```diff
@@ -0,0 +1,77 @@
...
+    MulticomponentData,
+    RawData,
+    Result,
+    WellList,
+)
+from allotropy.parsers.lines_reader import LinesReader
+from allotropy.parsers.utils.values import try_str_from_series_or_none
+
+
+def create_data(reader: LinesReader) -> Data:
```
nathan-stender, line 19: Please go ahead and put this as Data.create method in the structure file.  

This function is not over-large, and I think after this refactor there will be some other patterns we can identify to simplify further :D 

For future reference, if we were going to break up a file, I would lean towards just breaking up large Structure classes into their own files. But in this case, I think it's OK to be in a single file.

nathan-stender (changes requested):

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_data_creator.py
alejandro-salgado: I tried to move this function into a Data.create function. But I was not able to fix a circular reference issue that results.

The problem is that this method creates the calculated data documents and those use the other structure classes
So, If I move this code to the structure a circular reference is created.

nathan-stender: Ah ok, I thought it was just due to size. I still think we can make it work, but OK with merging this and taking a crack at it later. 👍

nathan-stender (approved):

nathan-stender:
Thanks again for all the hard working improving code consistency!

PR #163, opened 2023-11-30
Fix allotrope_from_io examples in README

bdworth:
`allotrope_from_io` has a required `file_name` parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,11 +41,13 @@ or, convert any IO:
 from allotropy.parser_factory import Vendor
 from allotropy.to_allotrope import allotrope_from_io
 
-with open("filename.txt") as f:
-    asm_schema = allotrope_from_io(f, Vendor.MOLDEV_SOFTMAX_PRO)
+file_name = "filepath.txt"
+
+with open(file_name) as f:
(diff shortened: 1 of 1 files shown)


nathan-stender (approved):

README.md
```diff
@@ -41,11 +41,13 @@ or, convert any IO:
...
 from allotropy.to_allotrope import allotrope_from_io
 
-with open("filename.txt") as f:
-    asm_schema = allotrope_from_io(f, Vendor.MOLDEV_SOFTMAX_PRO)
+file_name = "filepath.txt"
+
+with open(file_name) as f:
+    asm_schema = allotrope_from_io(f, file_name, Vendor.MOLDEV_SOFTMAX_PRO)
 
 bytes_io = BytesIO(file_stream)
```
mdlu, line 49: this example here is two separate examples, right? it isn't clear I think, maybe we can add some comments to clearly delineate the two

also this README doesn't feel quite right, what is `file_stream`? maybe this should be like
```
bytes_io = BytesIO(b'This is a sample file.')
```

mdlu (changes requested):

PR #171, opened 2023-12-06
refac: Simplify Moldev Softmax Pro parsing with dataclasses

alejandro-salgado:
- Simplify Moldev Softmax Pro parsing with dataclasses


tests/parsers/moldev_softmax_pro/testdata/expected_abs_endpoint.json
```diff
@@ -1,5484 +1,5484 @@
...
-          },
-          "processed data aggregate document": {
-            "processed data document": [
-              {
-                "data processing description": "processed data",
-                "processed data": 0.01
-              }
-            ]
-          }
+        "measurement identifier": "69b2d0aa-2a71-4998-b962-c95f9790f0d9",
```
nathan-stender, line 3: If this is just a refactor, then why did the expected test files change?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -189,20 +170,18 @@ def is_empty(self) -> bool:
...
 
 
-@dataclass(frozen=True)
-class PlateBlock(Block):
-    block_type: str
-    name: Optional[str]
-    export_format: Optional[str]
-    read_type: Optional[str]
-    data_type: Optional[str]
+@dataclass
```
nathan-stender, line 173: I don't believe I see anywhere where `PlateHeader` is modified, why did we remove `frozen=True`?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -105,18 +92,18 @@ class Block:
     raw_lines: list[str]
 
     @staticmethod
-    def create(lines: list[str]) -> Block:
+    def create(lines_reader: CsvReader) -> Block:
```
nathan-stender, line 95: Can we change `lines_reader` -> `csv_reader` or just `reader`

nathan-stender (changes requested):

bdworth (changes requested):
I haven't reviewed in detail, but I also wonder why test files need to change so dramatically for a refactor.

tests/parsers/moldev_softmax_pro/testdata/expected_abs_endpoint.json
slopez-b: @nathan-stender @bdworth After looking at the diffs of test files and running the unit tests on the old files this is what I found:

- `expected_lum_spectrum.json`: no significant changes on the file, rolled back to the previous version, all tests passed.
- `expected_abs_endpoint.json`: the type of the `dimensions` arrays items of the `data cube`s  changed from `int` to `str`. The `TDatacubeData` class defines the `dimenisions` property as a list of `TDimensionArray`, which accepts both `int` and `str`.
- `expected_fl_kinetic.json`: The problem here was that the `processed data` was being recorded with less significant figures when compared to the source file (ex. `0.014304468253968255` vs `0.0143044682539682`). It seems like the changes fixed this issue, but I haven't identified what was causing it.

Do you think these are good reasons to accept the changes to the test files?

nathan-stender: For `expected_abs_endpoint.json`, I think we should be getting the values back to `int`, the data cube values are permissive, but it's clear this data is really int from context.

For `expected_fl_kinetic.json`, if I read correctly we are increasing significant figures? (can't see the diff in github). If that is the case, I think this is OK

nathan-stender (changes requested):
Responded inline, but putting here to request changes, we should change the ints in the data cube back to int.

tests/parsers/moldev_softmax_pro/testdata/expected_abs_endpoint.json
LimulusPolyphemus: @nathan-stender @slopez-b, I would hesitant to too caught up on the change data cube in `expected_abs_endpoint.json` because the in the planned update to the Plate Reader ASM, this data cube will be removed in the endpoint read.

nathan-stender: Just pushed a fix

nathan-stender: For `expected_fl_kinetic.json`, the current implementation is dropping digits compared to the source data. I don't think this is acceptable, as this is scientific data.

nathan-stender: Found an fixed

slopez-b: @nathan-stender You are absolutely right. If mistakenly interpreted it the wrong way (I thought we were actually adding digits)

nathan-stender (approved):

slopez-b (approved):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,16 +1,15 @@
...
 from allotropy.parsers.moldev_softmax_pro.softmax_pro_structure import Data
 from allotropy.parsers.vendor_parser import VendorParser
 
 
 class SoftmaxproParser(VendorParser):
     def to_allotrope(self, named_file_contents: NamedFileContents) -> Any:
-        contents = named_file_contents.contents
-        lines_reader = LinesReader(contents)
-        data = Data.create(lines_reader)
+        reader = CsvReader(named_file_contents.contents, encoding=None)
```
bdworth, line 11: @alejandro-salgado , this changed functionality: passing `encoding=None` means we no longer assume UTF-8 and let chardet detect the encoding. Was this intended?

PR #174, opened 2023-12-07
Wrap pd.read_csv()

bdworth:
Might still want to do this to ensure a DataFrame is returned, but might not be as useful after #175


bdworth:
replaced by #266

PR #178, opened 2023-12-11
fix: Fix Quantstudio well calculated documents 

nathan-stender:
There is a bug in per-well calculated documents, where we assume there can only be one `quantity` calculated document per well.

This does not work when there are multiple targets in a well, because may each have a quantity value.

Fix this by making `Well.calculated_document` -> `Well.calculated_documents` and adjusting accordingly.

Added unit test that exercises this code.

Also fixed some error messages that similarly assumed only one well, making them less clear than they could be.


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_parser.py
```diff
@@ -99,7 +99,7 @@ def _get_model(self, data: Data, file_name: str) -> Model:
                     )
                     for well in data.wells
                 ],
-                calculated_data_aggregate_document=self.get_outter_calculated_data_aggregate_document(
+                calculated_data_aggregate_document=self.get_outer_calculated_data_aggregate_document(
```
bdworth, line 102: nice, but could we move this rename out of this fixing PR?

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -252,9 +252,9 @@ def create_generic(data: pd.Series[str]) -> WellItem:
 class Well:
     identifier: int
     items: dict[str, WellItem]
+    calculated_documents: list[CalculatedDocument]
```
bdworth, line 255: It seems like we only want to allow appending new documents after the class is instantiated, and that we always want to instantiate with an empty list. So how about:
`__calculated_documents: list[CalculatedDocument] = []`

... with a property a la how the two fields below are accessed, and a new method `add_calculated_document`?

Benefits:
1. Make it harder to accidentally reset `calculated_documents` 
2. No changing existing `Well` instantiations

bdworth (changes requested):
logic fix lgtm, just some structure comments

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_parser.py
nathan-stender: No

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
nathan-stender: Yeah, the reason I avoided this was because it's inadvisable to use default `[]`, and I didn't want to do the extra work of dealing with if null, etc, etc. But you're right it's better to be fully correct than mostly correct, fixing.

bdworth (changes requested):
just one more comment -- happy to approve shortly after addressed

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -285,6 +279,21 @@ def melt_curve_raw_data(self) -> Optional[MeltCurveRawData]:
     def melt_curve_raw_data(self, melt_curve_raw_data: MeltCurveRawData) -> None:
         self._melt_curve_raw_data = melt_curve_raw_data
 
+    @property
+    def calculated_documents(self) -> list[CalculatedDocument]:
+        return self._calculated_documents if self._calculated_documents else []
+
+    @calculated_documents.setter
```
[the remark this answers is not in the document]
nathan-stender: Done

bdworth (changes requested):
Looks like you conflict-bombed yourself with #179 merging

bdworth (approved):

PR #185, opened 2023-12-14
Improve SoftMaxPro Block building

bdworth:
Separate building duties from the classes they build for `Block` and its children.

Scoped to keep the git diff as small as possible -- so, e.g., the builder classes are near the classes they build.


nathan-stender (changes requested):
I don't think this is the direction we should be going here.

The goal of the structure/parser pattern in the parsers is for the structure classes to be a flat dataclass with a `create` method that reads the raw data. We have explicitly removed builder patterns from the codebase, because they just have a tendency to hide complexity that points to places where we are violating the pattern above.

There are problems with this parser in particular violating the structure/parser pattern (inherited from legacy implementation that existed before we started using this pattern). But the intended solution is to remove the parts of the code that are interacting with allotrope models (this should be done in the parser).

See the `aglient_gen5` adapter for an example of another legacy parser that we have done refactoring to remove the allotrope models from the structure classes (this one is still in progress as well, but it is closer)

bdworth:
Discussed offline, makes sense not to go this direction

PR #188, opened 2023-12-14
Test exception messages

bdworth:


tests/parsers/beckman_vi_cell_xr/vi_cell_xr_parser_test.py
```diff
@@ -36,5 +36,7 @@ def test_parse_vi_cell_xr_to_asm_expected_contents(output_file: str) -> None:
 
 def test_perse_vi_cell_xr_file_without_required_fields_then_raise() -> None:
     test_filepath = "tests/parsers/beckman_vi_cell_xr/testdata/v2.04/Beckman_Vi-Cell-XR_example02_instrumentOutput.xls"
-    with pytest.raises(AllotropeConversionError):
+    with pytest.raises(AllotropeConversionError) as exc_info:
```
nathan-stender, line 39: Why not use `match=` for this and all the cases below?

If it's the length of the string, you could define `expected` above this line and then use `match=expected`?

nathan-stender (changes requested):

tests/parsers/beckman_vi_cell_xr/vi_cell_xr_parser_test.py
bdworth: I generally prefer `match` except where I'd have to change characters for regex escaping. That's the case here and everywhere I tested the exact message below.

nathan-stender: Ah I see. You can use `re.escape()` on the string to handle that, but I'm fine either way.

nathan-stender (approved):

tests/parsers/beckman_vi_cell_xr/vi_cell_xr_parser_test.py
bdworth: Ah, right. I'll use `re.escape()`.

nathan-stender (approved):

PR #196, opened 2023-12-15
tests: Improve schema generation test

nathan-stender:
Add the ability to run generate schemas script as a dry run and return the number of models changed. 

Change the test to validate that no model files are changed.

This improves test coverage and prevents the test from making changes to files.

### DIFF
--- a/src/allotropy/allotrope/schema_parser/generate_schemas.py
+++ b/src/allotropy/allotrope/schema_parser/generate_schemas.py
@@ -70,6 +70,7 @@ def generate_schemas(root_dir: Path, *, dry_run: Optional[bool] = False) -> list
     """Generate schemas from JSON schema files.
 
     :root_dir: The root directory of the project.
+    :dry_run: If true, does not save changes to any models, but still returns the list of models that would change.
     :return: A list of model files that were changed.
     """
 
--- a/src/allotropy/allotrope/schema_parser/generate_schemas.py
+++ b/src/allotropy/allotrope/schema_parser/generate_schemas.py
@@ -3,6 +3,7 @@ import os
 from pathlib import Path
 import re
 import subprocess  # noqa: S404, RUF100
+from typing import Optional
 
 from autoflake import fix_file  # type: ignore[import-untyped]
 from datamodel_code_generator import (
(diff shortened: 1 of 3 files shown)


tests/allotrope/schema_parser/generate_schemas_test.py
```diff
@@ -5,4 +5,7 @@
 
 def test_generate_schemas_runs_to_completion() -> None:
     root_dir = Path(__file__).parent.parent.parent.parent
-    assert generate_schemas(root_dir) == 10
```
mdlu, line 8: not having to update this int every time anymore is huge 👏

src/allotropy/allotrope/schema_parser/generate_schemas.py
```diff
@@ -65,17 +66,17 @@ def files_equal(path1: str, path2: str) -> bool:
     return True
 
 
-def generate_schemas(root_dir: Path) -> int:
+def generate_schemas(root_dir: Path, *, dry_run: Optional[bool] = False) -> list[str]:
```
mdlu, line 69: nit: could you add a description of what `dry_run` is? also, the name was a little confusing to me, maybe like `validate` or `validation_run` or something like that

tests/allotrope/schema_parser/generate_schemas_test.py
```diff
@@ -5,4 +5,7 @@
...
+    models_changed = generate_schemas(root_dir, dry_run=True)
+    assert (
+        not models_changed
+    ), f"Expected no models files to have changed by generate-schemas script, found changes in: {models_changed}. Please run scripts:generate-schemas and validate the changes."
```
mdlu, line 11: maybe we can have the full line to run (`hatch run scripts:generate-schemas`) in the error message so people can just copy-paste?

src/allotropy/allotrope/schema_parser/generate_schemas.py
nathan-stender: Added a description for `dry_run`. `dry_run` is pretty commonly used for scripts/commands to mean run without saving changes, so I'm inclined to keep the name.

tests/allotrope/schema_parser/generate_schemas_test.py
nathan-stender: Done!

mdlu (approved):

PR #197, opened 2023-12-15
Test allotrope_model_from_file()

bdworth:
Test a previously unused function. (It's used in neither src nor tests.)

### DIFF
--- a/tests/parsers/agilent_gen5/to_allotrope_test.py
+++ b/tests/parsers/agilent_gen5/to_allotrope_test.py
@@ -49,8 +49,8 @@ def test_to_allotrope_absorbance(filename: str) -> None:
 
 
 # Test allotrope_model_from_file().
-@pytest.mark.parametrize("filename", ABSORBENCE_FILENAMES)
-def test_model_from_file_absorbance(filename: str) -> None:
+def test_model_from_file_absorbance() -> None:
+    filename = ABSORBENCE_FILENAMES[0]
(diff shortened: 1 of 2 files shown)


tests/parsers/agilent_gen5/to_allotrope_test.py
```diff
@@ -1,35 +1,64 @@
...
     )
     allotrope_dict = from_file(test_filepath, VENDOR_TYPE)
-    validate_schema(
-        allotrope_dict,
-        "ultraviolet-absorbance/BENCHLING/2023/09/ultraviolet-absorbance.json",
+    _validate_allotrope_dict(allotrope_dict, expected_filepath)
+
+
+# Test allotrope_model_from_file().
+@pytest.mark.parametrize("filename", ABSORBENCE_FILENAMES)
```
nathan-stender, line 52: `allotrope_model_from_file` is tested implicitly by calls to `allotrope_from_io`. 

I'm fine with adding a test for it if you really want, but I don't think we need to run on all test cases here.

nathan-stender (changes requested):

tests/parsers/agilent_gen5/to_allotrope_test.py
bdworth: Updated PR description. The function currently has no calls, so I figured it should be tested. Fine just running on one test case.

nathan-stender (approved):

PR #201, opened 2023-12-16
Addition of SquareCentimetersPerGram to units.json

LimulusPolyphemus:
Addition of SquareCentimetersPerGram to support use of `attenuation coefficient` in `spectrophotometry` schema


nathan-stender (changes requested):
As with other PRs, you will need to add py models as well

nathan-stender (approved):

PR #202, opened 2023-12-16
Addition of tQuantityValueSquareCentimetersPerGram to custom.json

LimulusPolyphemus:
Addition of tQuantityValueSquareCentimetersPerGram to custom.json to support user in `spectrophotometry` schema


nathan-stender (changes requested):
As with other PRs, you will need to add py models as well

LimulusPolyphemus:
@nathan-stender I believe this check is failing because it dependent on the change in #201, once that PR merges, this should pass as well.

nathan-stender (approved):

PR #203, opened 2023-12-16
Additional mass concentration units for units.json

LimulusPolyphemus:
Adding ng/uL, ug/uL, ng/mL, ug/mL, and mg/mL mass concentration units to units.json.


nathan-stender (commented):
You will need to add the models in `src/allotropy/allotrope/models/shared/definitions/units.py` as well

nathan-stender (approved):

PR #204, opened 2023-12-16
Additional mass concentration tQuantityValue for custom.json

LimulusPolyphemus:
Addition of modified tQuantityValue for mass concentations: ng/uL, ug/uL, ng/mL, ug/mL, mg/mL


nathan-stender (changes requested):
You will need to add these in `src/allotropy/allotrope/models/shared/definitions/custom.py` as well

nathan-stender (approved):

PR #209, opened 2023-12-19
Chore: Adding more endpoint test data for moldev_softmax_pro

LimulusPolyphemus:
Adding test files from SoftMax Pro for endpoint reads for Absorbance, Fluorescence, and Luminescence measurements containing multiple plates, multiple measurements, and group data.

These files are in their native encoding exported from SoftMax Pro version 7.1.2 running on machine with Windows 10 Enterprise.


bdworth (approved):
Before merging, could you please call out in the description that these files are:
1. from SoftMaxPro
2. uploaded as is in their native encoding that isn't UTF-8
?

PR #210, opened 2023-12-19
feat: non numeric results tquatity value

slopez-b:
Some of the adapters need a way to represent measurements that are either incorrect, or fall outside the defined ranges of the instruments. Currently, these properties in ASM only accept numeric values. This PR updates the `tQuantityValue` definition to support constants for non-numeric values.

The accepted values are:

- "NaN"
- "+Infinity"
- "-Infinity"


src/allotropy/allotrope/models/shared/definitions/definitions.py
```diff
@@ -52,9 +52,15 @@ class TStatisticDatumRole(Enum):
     minimum_value_role = "minimum value role"
 
 
+class TNonNumericQuantityValue(Enum):
+    nan = "NaN"
+    plus_infinite = "+Infinity"
+    minus_infinite = "-Infinity"
+
```
nathan-stender, line 59:
```suggestion
    NaN = "NaN"
    Infinity = "Infinity"
    field_Infinity = "-Infinity"
```

nathan-stender (changes requested):

src/allotropy/allotrope/models/shared/definitions/definitions.py
nathan-stender: Requesting this change because this is how the schema generator names these values, want to keep as close to auto-generated as possible.

src/allotropy/allotrope/schemas/shared/definitions/definitions.json
```diff
@@ -132,7 +132,20 @@
         "type": "object",
         "properties": {
             "value": {
-                "type": "number"
+                "oneOf": [
```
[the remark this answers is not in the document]
nathan-stender: Requesting this change because it causes the schema generator to actually generate the model we want, "const" does not

src/allotropy/allotrope/models/shared/definitions/definitions.py
slopez-b: @nathan-stender @nathan-stender Is it ok to change `+Infinity` to just `Infinity?

keeping the `+Infinity` I got:
```python
class ValueEnum(Enum):
    NaN = "NaN"
    field_Infinity = "+Infinity"
    field_Infinity_1 = "-Infinity"
```

nathan-stender (approved):

PR #213, opened 2023-12-20
Chemometic Nucleoview: support for non-numeric values

bwolfe-benchling:
Add support for non-numeric values in fields that are mapped to TQuantity type values.

### DIFF
--- a/src/allotropy/parsers/chemometec_nucleoview/nucleoview_reader.py
+++ b/src/allotropy/parsers/chemometec_nucleoview/nucleoview_reader.py
@@ -34,5 +34,10 @@ class NucleoviewReader:
                 raw_data["Date time"] + raw_data["Time zone offset"]
             )
         raw_data["Sample ID"] = raw_data["Image"].str.split("-", n=3).str[3]
+        raw_data.rename(
+            {"Estimated cell diameter [um]": "Estimated cell diameter (um)"},
+            axis=1,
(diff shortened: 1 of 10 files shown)


src/allotropy/parsers/chemometec_nucleoview/nucleoview_reader.py
```diff
@@ -34,5 +34,10 @@ def read(cls, contents: IOType) -> pd.DataFrame:
                 raw_data["Date time"] + raw_data["Time zone offset"]
             )
         raw_data["Sample ID"] = raw_data["Image"].str.split("-", n=3).str[3]
+        raw_data.rename(
+            {"Estimated cell diameter [um]": "Estimated cell diameter (um)"},
+            axis=1,
+            inplace=True,
+        )
```
slopez-b, line 41: Please avoid using `inplace=True`, it is considered to be potentially harmful and there are even [plans to deprecate it in the future](https://github.com/pandas-dev/pandas/issues/16529).
```suggestion
        raw_data = raw_data.rename(
            {"Estimated cell diameter [um]": "Estimated cell diameter (um)"},
            axis=1,
        )
```

slopez-b (changes requested):
Looks good, only one small change requested.

slopez-b (dismissed):

slopez-b (approved):

PR #215, opened 2023-12-22
Fix: calculated data name in elmer envision

slopez-b:
In the Perkin Elmer Envision Adapter the calculated data name of the calculated data documents was being incorrectly parsed.
- Instead of grabbing the left part of the `formula` column on the `Plate information` block, the adapter was taking the right part.
- Added additional test for the structure class in charge of parsing this property.

### DIFF
--- a/tests/parsers/perkin_elmer_envision/testdata/PE_Envision_fluorescence_example01.json
+++ b/tests/parsers/perkin_elmer_envision/testdata/PE_Envision_fluorescence_example01.json
@@ -5490,7 +5490,7 @@
         "calculated data aggregate document": {
             "calculated data document": [
                 {
-                    "calculated data name": "(X / Y) * Z where X = AC HTRF Laser [Eu](1) channel 1 window 1  Y = AC HTRF Laser [Eu](1) channel 2 window 1  Z = 10000",
+                    "calculated data name": "Calc 1: General",
                     "calculated result": {
                         "value": 3912.12920565,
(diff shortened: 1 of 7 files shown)


tests/parsers/perkin_elmer_envision/testdata/PE_Envision_fluorescence_example01.json
```diff
@@ -5485,1019 +5485,1019 @@
...
             "software version": "1.14.3049.1193",
             "ASM converter name": "allotropy",
-            "ASM converter version": "0.1.10"
+            "ASM converter version": "0.1.12"
         },
         "calculated data aggregate document": {
             "calculated data document": [
                 {
-                    "calculated data name": "(X / Y) * Z where X = AC HTRF Laser [Eu](1) channel 1 window 1  Y = AC HTRF Laser [Eu](1) channel 2 window 1  Z = 10000",
+                    "calculated data name": "Calc 1: General",
```
nathan-stender, line 5493: it seems like you could find/replace "calculated data name" value in the test files and prevent the "measurement identifier" changes, making the test changes clearer, can you do that please?

nathan-stender (changes requested):
Change looks good, but I'd like to avoid just re-generating the test files whenever possible. It seems like it should be possible in this case, so requesting you do that instead.

alejandro-salgado (approved):

nathan-stender (approved):

PR #216, opened 2024-01-02
feat: Add calculated data documents to Softmax Pro

alejandro-salgado:
- Add calculated data documents to Softmax Pro


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,349 @@
...
         data = Data.create(reader)
-        return self._get_model(data)
+        return self._get_model(named_file_contents.original_file_name, data)
+
+    def _get_model(self, file_name: str, data: Data) -> Model:
+        return Model(
+            field_asm_manifest="http://purl.allotrope.org/json-schemas/adm/plate-reader/BENCHLING/2023/09/plate-reader.schema",
+            plate_reader_aggregate_document=PlateReaderAggregateDocument(
+                device_system_document=DeviceSystemDocument(
+                    device_identifier="null",
```
oxytocinlove, line 71: should these be None instead of the string `"null"`?

oxytocinlove (dismissed):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
alejandro-salgado: The device identifier attribute is mandatory. The null value is a placeholder because there is no information in the file to get a proper device identifier.

bdworth:
I just made the change that I requested above.

tests/parsers/moldev_softmax_pro/testdata/abs_endpoint_plates.json
```diff
@@ -11,7 +11,7 @@
                     },
                     "measurement document": [
                         {
-                            "measurement identifier": "e3e8807d-1f02-4a4f-80be-5398a7f4235d",
+                            "measurement identifier": "d380f805-450a-4120-a09d-fee2ff7710ac",
```
bdworth, line 14: why did all these measurement identifiers change?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -71,10 +80,14 @@ def _get_model(self, file_name: str, data: Data) -> Model:
                 ),
                 plate_reader_document=[
                     self._get_plate_reader_document_item(plate_block, position)
-                    for plate_block in data.get_plate_block()
+                    for plate_block in data.block_list.plate_blocks.values()
                     for position in plate_block.iter_wells()
                 ],
-                calculated_data_aggregate_document=None,
+                calculated_data_aggregate_document=CalculatedDataAggregateDocument(
+                    calculated_data_document=list(
```
bdworth, line 87: Similar to comments in #217, let's have the helper return a list rather than an Iterator unless you see a near-term future need for an iterator.

P.S. Although in this case, your get-iterator method has more than one `yield`, so maybe it's fine as is?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -144,21 +157,23 @@ def _iter_fluorescence_measurement_document(
...
+                measurement_identifier=data_element.uuid,
                 fluorescence=TRelativeFluorescenceUnit(value=data_element.value),
                 compartment_temperature=(
                     None
                     if data_element.temperature is None
-                    else TQuantityValueDegreeCelsius(data_element.temperature)
+                    else TQuantityValueDegreeCelsius(
+                        ValueEnum.NaN
+                        if math.isnan(data_element.temperature)
+                        else data_element.temperature
```
bdworth, line 170: Please create a helper for this with-Nan-value construction and use it here and below.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -264,3 +287,63 @@ def _iter_absorbance_measurement_document(
...
+                        unit="unitless",
+                        value=reduced_data_element.value,
+                    ),
+                    data_source_aggregate_document=DataSourceAggregateDocument1(
+                        data_source_document=[
+                            DataSourceDocumentItem(
+                                data_source_identifier=w.uuid,
+                                data_source_feature=plate_block.plate_block_type,
+                            )
+                            for w in plate_block.iter_data_elements(
```
bdworth, line 309: could you use a more descriptive variable name than `w`?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -264,3 +287,63 @@ def _iter_absorbance_measurement_document(
...
+                            )
+                        ]
+                    ),
+                )
+
+        for group_block in data.block_list.group_blocks:
+            for group_sample_data in group_block.group_data.sample_data:
+                for group_data_element in group_sample_data.data_elements:
+                    plate_block = data.block_list.plate_blocks[group_data_element.plate]
+                    for entrie in group_data_element.entries:
```
bdworth, line 320: s/entrie/entry

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -264,3 +287,63 @@ def _iter_absorbance_measurement_document(
...
             )
+
+    def _iter_calculated_data_documents(
+        self, data: Data
+    ) -> Iterator[CalculatedDataDocumentItem]:
+        for plate_block in data.block_list.plate_blocks.values():
+            for reduced_data_element in plate_block.iter_reduced_data():
+                yield CalculatedDataDocumentItem(
+                    calculated_data_identifier=str(uuid.uuid4()),
+                    calculated_data_name="Reduced",
```
bdworth, line 298: Will we be using "Reduced" in other parsers? If so, should we make a constant for this?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -264,3 +287,63 @@ def _iter_absorbance_measurement_document(
...
+                    calculated_result=TQuantityValue(
+                        unit="unitless",
```
bdworth, line 300: should "unitless" be a library-wide constant? Seems like it's used in other parsers.

bdworth (changes requested):
a few comments. Also worth noting: there are a lot of unresolved comments on the parent PR, so as it changes this might have to change.

tests/parsers/moldev_softmax_pro/testdata/abs_endpoint_plates.json
alejandro-salgado: They change because I update the expected output using a run of the new code. I did this because it was too long to generate by hand, however I went through every change making sure that only the expected values were changed.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
alejandro-salgado: There are two types of calculated data documents, the first one is from the reduced data in the plate block and the second one is from the group block. The function iter_calculated_data_documents iterates over all of them, that is why is an iterator. However, It can also be implemented as a list

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
alejandro-salgado: Reduced is specific to this parser. I put a constant for this value.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -21,13 +22,53 @@
...
+    try_str_from_series,
+    try_str_from_series_or_none,
 )
 
 BLOCKS_LINE_REGEX = r"^##BLOCKS=\s*(\d+)$"
 END_LINE_REGEX = "~End"
 EXPORT_VERSION = "1.3"
 
 
+def is_float(value: Any) -> bool:
```
bdworth, line 34: again, this is confusingly named: 
1. I'd expect `is_float(float("NaN"))` to return True. How about `is_non_nan_float`?
2. This will return `true` for any String that can parse as a float, but I wouldn't expect `is_float('3')` to return True.

How about `can_parse_as_float_non_nan` or similar?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -21,13 +22,53 @@
...
+    try:
+        number = float(value)
+        return False if math.isnan(number) else True
+    except Exception:
+        return False
```
bdworth, line 39: catching `Exception` is generally a bad practice, as is making try blocks too large. How about instead:

```suggestion
    try:
        number = float(value)
    except ValueError:
        return False
    return not math.isnan(number)
```

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -21,13 +22,53 @@
...
+
+
+def rm_df_columns(data: pd.DataFrame, pattern: str) -> pd.DataFrame:
+    new_data = data
+    for column in data.columns:
+        if re.match(pattern, column):
+            new_data = new_data.drop(columns=[column])
+    return new_data
```
bdworth, line 47: Let's only call `drop` once:

```suggestion
    columns_to_drop = []
    for column in data.columns:
        if re.match(pattern, column):
            columns_to_drop.append(column)
    return data.drop(columns=columns_to_drop)
```

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -21,13 +22,53 @@
...
+
+
+def try_str_from_series_multikey_or_none(
+    data: pd.Series,  # type: ignore[type-arg]
```
bdworth, line 51: better `pd.Series[Any]` than suppressing, here and below

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -21,13 +22,53 @@
...
+    possible_keys: list[str],
```
bdworth, line 52: seems like `possible_keys` should be `set` here and below

bdworth (changes requested):
High-level comment in addition to line-level comments: there are no calls to `math.isnan()` before this PR and currently 7 after. Could you please consolidate the nan logic a bit?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -45,9 +61,9 @@
 )
 from allotropy.parsers.vendor_parser import VendorParser
 
-DEVICE_TYPE = "plate reader"
-EPOCH = "1970-01-01T00:00:00-00:00"
-NULL = "null"
+
+def float_or_nan(value: float) -> Union[ValueEnum, float]:
```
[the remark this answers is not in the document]
bdworth: P.S. I'm trying to clear up this confusion in #224

src/allotropy/parsers/lines_reader.py
```diff
@@ -40,8 +40,14 @@ def __init__(self, lines: list[str]) -> None:
         self.lines = lines
         self.current_line = 0
 
+    def line_exists(self, line: int) -> bool:
```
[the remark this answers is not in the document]
alejandro-salgado: The idea is to allow parsers to consult if a specific line exists.

bdworth (changes requested):
Please merge `main` into `softmax-plate-reader`, then `softmax-plate-reader` into here.

bdworth (changes requested):
just a few refactors around NaN, and then we're done!

bdworth (approved):
Looks good! Could you please clean up the PR description before merging (or just edit the commit description when merging)?

PR #217, opened 2024-01-02
feat: Update plate reader schema in Moldev Softmax Pro

alejandro-salgado:
- Update plate reader schema in Moldev Softmax Pro


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
-        return self._get_model(data)
+        return self._get_model(named_file_contents.original_file_name, data)
+
+    def _get_model(self, file_name: str, data: Data) -> Model:
+        return Model(
+            field_asm_manifest="http://purl.allotrope.org/json-schemas/adm/plate-reader/BENCHLING/2023/09/plate-reader.schema",
+            plate_reader_aggregate_document=PlateReaderAggregateDocument(
+                device_system_document=DeviceSystemDocument(
+                    device_identifier="null",
+                    model_number="null",
```
bdworth, line 63: I really don't think we should be using a hardcoded "null" for either of these non-nullable identifiers. If we're saying that it's ok for a `DeviceSystemDocument` to be missing these fields, then they should allow `None` as an option. If it's not ok to be missing these fields for other uses of `DeviceSystemDocument`, then we may need a slightly different dataclass here.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                ),
+                data_system_document=DataSystemDocument(
+                    file_name=file_name,
+                    software_name="SoftMax Pro",
+                    software_version=None,
```
bdworth, line 68: software_version is `None` by default, so no need to set here.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                    ASM_converter_name=ASM_CONVERTER_NAME,
+                    ASM_converter_version=ASM_CONVERTER_VERSION,
+                ),
+                plate_reader_document=[
+                    self._get_plate_reader_document_item(plate_block, position)
+                    for plate_block in data.get_plate_block()
+                    for position in plate_block.iter_wells()
+                ],
+                calculated_data_aggregate_document=None,
```
bdworth, line 77: again, no need to set to None when it's the default

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+            ),
+        )
+
+    def _get_plate_reader_document_item(
+        self, plate_block: PlateBlock, position: str
+    ) -> PlateReaderDocumentItem:
+        if plate_block.plate_block_type == "Absorbance":
```
bdworth, line 84: extract `plate_block.plate_block_type` to a local var please -- it's used 3X

(If we could require Python >=3.10, we could use `case`, but we have to support 3.9 for now)

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                    plate_well_count=TQuantityValueNumber(plate_block.header.num_wells),
+                    container_type=ContainerType.well_plate,
+                    measurement_document=list(
+                        self._iter_luminescence_measurement_document(
+                            plate_block, position
+                        ),
+                    ),
+                )
+            )
+        elif plate_block.plate_block_type == "Fluorescence":
```
bdworth, line 110: There's a LOT of repeated code in these three blocks. The only differences are in which `self._iter...` method is called. Could you please refactor to create the `measurement_document` in each path, then create the `PlatereaderDocumentItem` with that?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+            return PlateReaderDocumentItem(
+                measurement_aggregate_document=MeasurementAggregateDocument(
+                    measurement_time=EPOCH,
+                    plate_well_count=TQuantityValueNumber(plate_block.header.num_wells),
+                    container_type=ContainerType.well_plate,
+                    measurement_document=list(
```
bdworth, line 116: I don't see the point in making the helper methods return `Iterator`s if the only places they're called will immediately create lists from the iterators. How about just making the helpers return lists?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                        self._iter_fluorescence_measurement_document(
+                            plate_block, position
+                        )
+                    ),
+                )
+            )
+        else:
+            error = "Unable to find valid plate block type."
```
bdworth, line 124: Could you include the invalid `plate_block.plate_block_type` in this message?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+            raise AllotropeConversionError(error)
+
+    def _iter_fluorescence_measurement_document(
+        self, plate_block: PlateBlock, position: str
+    ) -> Iterator[FluorescencePointDetectionMeasurementDocumentItems]:
```
bdworth, line 129: see note above -- this and other similar methods should return `list` unless you anticipate other uses where you won't immediately be copying into lists.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+            scan_position = (
+                ScanPositionSettingPlateReader.bottom_scan_position__plate_reader_
+            )
+        elif plate_block.header.scan_position == "FALSE":
+            scan_position = (
+                ScanPositionSettingPlateReader.top_scan_position__plate_reader_
+            )
+        else:
+            error = "Unable to find valid scan position."
+            raise AllotropeConversionError(error)
```
bdworth, line 140: Could you extract this block to a helper -- e.g., `_get_fluorescence_plate_block_scan_position`?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+
+        reads_per_well = assert_not_none(
```
bdworth, line 142: `reads_per_well` can't be `None`

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                    sample_identifier=f"{plate_block.header.name} {data_element.position}",
+                    sample_role_type=None,
+                    well_location_identifier=None,
+                    vial_location_identifier=None,
+                    mass_concentration=None,
+                ),
+                device_control_aggregate_document=UltravioletAbsorbancePointDetectionDeviceControlAggregateDocument(
+                    device_control_document=[
+                        UltravioletAbsorbancePointDetectionDeviceControlDocumentItem(
+                            device_type="plate reader",
```
bdworth, line 258: could you extract a constant rather than using "plate reader" 3X?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                    else TQuantityValueDegreeCelsius(data_element.temperature)
+                ),
+                sample_document=SampleDocument(
+                    location_identifier=data_element.position,
+                    well_plate_identifier=plate_block.header.name,
+                    sample_identifier=f"{plate_block.header.name} {data_element.position}",
+                    sample_role_type=None,
+                    well_location_identifier=None,
+                    vial_location_identifier=None,
+                    mass_concentration=None,
```
bdworth, line 253: again, no need to set =None if that's the default value

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,266 @@
...
+                fluorescence=TRelativeFluorescenceUnit(value=data_element.value),
+                compartment_temperature=(
+                    None
+                    if data_element.temperature is None
+                    else TQuantityValueDegreeCelsius(data_element.temperature)
+                ),
+                sample_document=SampleDocument(
+                    location_identifier=data_element.position,
+                    well_plate_identifier=plate_block.header.name,
+                    sample_identifier=f"{plate_block.header.name} {data_element.position}",
```
bdworth, line 161: general comment for this and following methods: `plate_block.header` is used a lot. consider extracting local vars for readability.

bdworth (changes requested):
Some initial comments after reading through the parser changes. I haven't reviewed `softmax_pro_structure` yet.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -445,55 +369,53 @@ def create(
...
+                raw_lines=reader.lines,
+                header=header,
+                block_data=TimeData.create(reader, header),
             )
         elif header.export_format == ExportFormat.PLATE_FORMAT.value:
-            PlateBlock._parse_plate_format_data(
-                header,
-                well_data,
-                plate_data=PlateData.create(reader, header),
+            return cls(
```
bdworth, line 409: out of scope for fixing in this PR, but having a static method in a parent class to potentially instantiate children is pretty ugly. I think this is a known issue though.

tests/parsers/moldev_softmax_pro/testdata/expected_abs_endpoint.json
bdworth, line 1: It makes me a little nervous to delete existing tests when making as many src changes as we're making in this PR. Is it possible to keep this existing test data and test it as it was before? And same question for the other two deleted test files.

tests/parsers/moldev_softmax_pro/to_allotrope_test.py
```diff
@@ -10,33 +10,26 @@
...
-        ),
-        (
-            "tests/parsers/moldev_softmax_pro/testdata/lum_spectrum_columns.txt",
-            "luminescence/BENCHLING/2023/09/luminescence.json",
-            "tests/parsers/moldev_softmax_pro/testdata/expected_lum_spectrum.json",
-        ),
-        (
-            "tests/parsers/moldev_softmax_pro/testdata/abs_endpoint_plates.txt",
-            "ultraviolet-absorbance/BENCHLING/2023/09/ultraviolet-absorbance.json",
-            "tests/parsers/moldev_softmax_pro/testdata/expected_abs_endpoint.json",
```
bdworth, line 28: reiterating comment from deleted test files: can we keep the existing test data without deleting it and run tests as before? Or will they no longer pass?

bdworth (changes requested):
finished first review pass

tests/parsers/moldev_softmax_pro/testdata/expected_abs_endpoint.json
alejandro-salgado: These files are no longer supported, that is why they are not included in the test any more.

tests/parsers/moldev_softmax_pro/to_allotrope_test.py
alejandro-salgado: A change in the plate reader schema generates a change in the output, so all tests must be updated

src/allotropy/parsers/lines_reader.py
```diff
@@ -40,8 +40,14 @@ def __init__(self, lines: list[str]) -> None:
...
         self.current_line = 0
 
+    def line_exists(self, line: int) -> bool:
+        return 0 <= line < len(self.lines)
+
     def current_line_exists(self) -> bool:
-        return 0 <= self.current_line < len(self.lines)
+        return self.line_exists(self.current_line)
+
+    def get_line(self, line: int) -> Optional[str]:
```
bdworth, line 49: This method is unused -- please delete.

And if you're deleting this method, then there's really no need to factor out `line_exists` here, so I think you can undo all the changes in this file.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -1,17 +1,267 @@
...
+    ScanPosition,
+)
+from allotropy.parsers.utils.values import (
+    assert_not_none,
+)
 from allotropy.parsers.vendor_parser import VendorParser
 
+DEVICE_TYPE = "plate reader"
+EPOCH = "1970-01-01T00:00:00-00:00"
+NULL = "null"
```
bdworth, line 50: Could you please add comments explaining why EPOCH and NULL need to exist? (I think it's because a time and string are required by the ASM.)

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -87,30 +45,24 @@ class DataType(Enum):
...
-            "Plate": PlateBlock,
-        }
-
-        for key, cls in block_cls_by_type.items():
-            if reader.match(f"^{key}"):
-                return cls.create(reader)
-
-        error = f"Expected block '{reader.get()}' to start with one of {sorted(block_cls_by_type.keys())}."
-        raise AllotropeConversionError(error)
+        raise NotImplementedError
```
bdworth, line 61: Let's delete this method as well as `NoteBlock.create()`. It's no longer implemented by `PlateBlock` or any of its children, it's only called as `GroupBlock.create()`, and `NoteBlock.create()` isn't called.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -180,52 +99,49 @@ class PlateHeader:
...
     num_rows: int
     excitation_wavelengths: Optional[list[int]]
     cutoff_filters: Optional[list[int]]
 
 
 @dataclass
-class WavelengthElement:
-    row: str
-    col: str
+class DataElement:
```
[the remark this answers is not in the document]
bdworth: Could you please freeze this class? It's safe to do so.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -518,198 +398,60 @@ def get_plate_block_cls(header_series: pd.Series[str]) -> type[PlateBlock]:
...
-                "data type", header.data_type, DataType._member_names_
-            )
-            raise AllotropeConversionError(msg)
-
-    @staticmethod
-    def split_wavelengths(values: Optional[str]) -> Optional[list[int]]:
-        return None if values is None else [int(v) for v in values.split()]
+    @classmethod
+    def get_plate_block_type(cls) -> str:
+        raise NotImplementedError
```
bdworth, line 403: Could you add `@abstractmethod` here and `ABC` to `PlateBlock`?

bdworth (changes requested):
looking better! Just a few small things to fix, then we should be good to merge.

bdworth (approved):

PR #220, opened 2024-01-05
Make SMP create() methods create.

bdworth:
A static initializer should initialize a class instance.

I'm also open to renaming these `create` methods to make it obvious that they don't always create but think that keeping the `create` naming consistent is nice.


bdworth:
done in #217

PR #223, opened 2024-01-10
Create try_str_from_series_or_default

bdworth:
Simplify a confusing signature by creating a new function. If a non-null default is supplied, there's no way to return `None`.

Change all callers that used the `default` param.


src/allotropy/parsers/utils/values.py
```diff
@@ -76,12 +76,20 @@ def assert_not_empty_df(df: pd.DataFrame, msg: str) -> pd.DataFrame:
...
+
+
 def try_str_from_series_or_none(
     data: pd.Series[Any],
     key: str,
-    default: Optional[str] = None,
 ) -> Optional[str]:
-    value = data.get(key, default)
+    value = data.get(key)
     return None if value is None else str(value)
```
mdlu, line 93: not a big deal but this could just be `return try_str_from_series_or_default(data, key, None)` now, right?

mdlu (approved):

src/allotropy/parsers/utils/values.py
bdworth: No, because `default` can't be `None` (and thus the return value can't be `None`).

PR #228, opened 2024-01-11
Extract HasUnit class

bdworth:
Make a parent class.


nathan-stender (approved):

bdworth:
merged as part of #243

PR #230, opened 2024-01-11
Nanodrop eight support

bwolfe-benchling:
Provide support for Nanodrop Eight


src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,271 @@
...
+    TQuantityValueMilliAbsorbanceUnit,
+    TQuantityValueMilligramPerMilliliter,
+    TQuantityValueNanogramPerMicroliter,
+    TQuantityValueNanogramPerMilliliter,
+    TQuantityValueNanometer,
+    TQuantityValuePicogramPerMilliliter,
+)
+from allotropy.allotrope.models.shared.definitions.definitions import (
+    TQuantityValue,
+    ValueEnum,
```
bdworth, line 18: ValueEnum has been renamed to InvalidJsonFloat

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,271 @@
...
+def _get_str(data_frame: pd.DataFrame, row: int, column: str) -> Optional[str]:
+    if column not in data_frame.columns:
+        return None
+    if pd.isna(data_frame.iloc[row][column]):
+        return None
+
+    return str(data_frame.iloc[row][column])
+
+
+def _get_float(data_frame: pd.DataFrame, row: int, column: str) -> float | ValueEnum:
```
bdworth, line 64: you can use `JsonFloat` instead of `float | ValueEnum`, here and below. Also note that this is still using Python 3.9 syntax, so if you want a union you need typing.Union.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,271 @@
...
+    try:
+        return float(data_frame.iloc[row][column])
+    except (ValueError, TypeError):
+        return ValueEnum.NaN
+
+
+def _get_concentration(conc: float | ValueEnum, unit: Optional[str]) -> Optional[T]:
+    if unit in CONCENTRATION_UNIT_TO_TQUANTITY and isinstance(conc, float):
+        cls = CONCENTRATION_UNIT_TO_TQUANTITY[unit]
+        return cls(value=conc)  # type: ignore[return-value]
```
bdworth, line 74: We should be able to remove this ignore. I'll get back to you on this.

bdworth (changes requested):

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bdworth: Eh, it's not going to be possible to remove this suppression without some larger refactoring of type classes. I'm fine leaving this in for now.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,272 @@
...
+}
+
+
+def _get_str(data_frame: pd.DataFrame, row: int, column: str) -> Optional[str]:
+    if column not in data_frame.columns:
+        return None
+    if pd.isna(data_frame.iloc[row][column]):
+        return None
+
+    return str(data_frame.iloc[row][column])
```
bdworth, line 62: could you extract `data_frame.iloc[row][column]` to a local var?

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,272 @@
...
+                    ASM_converter_version=ASM_CONVERTER_VERSION,
+                ),
+                device_system_document=DeviceSystemDocument(
+                    model_number="Nanodrop Eight",
+                    device_identifier="Nanodrop",
+                ),
+            ),
+        )
+
+    def _add_measurement_uuids(self, data: pd.DataFrame) -> pd.DataFrame:
```
bdworth, line 107: Since we're modifying `data` rather than creating a copy, I'd prefer not to have a return value.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,272 @@
...
+            calculated_data_name="A260/280",
+            calculated_result=TQuantityValue(
+                value=_get_float(data, row, "260/280"), unit="unitless"
+            ),
+            calculated_data_identifier=str(uuid.uuid4()),
+            data_source_aggregate_document=DataSourceAggregateDocument1(
+                [
+                    DataSourceDocumentItem(
+                        data_source_feature="absorbance",
+                        data_source_identifier=str(_get_str(data, row, "A260 uuid")),
```
bdworth, line 145: I'm skeptical of every instance of `str(_get_str(`, both from a naming perspective and looking at the impl.

Calling `str` on `None` returns `'None'`, which I really don't think is what we want.

For fields like `data_source_identifier` that require a `str`, we need to make sure we supply one. It seems like it should always exist since the columns were added though.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,272 @@
...
+    "ug/ul": TQuantityValueMicrogramPerMicroliter,
+    "ug/ml": TQuantityValueMicrogramPerMilliliter,
+    "mg/ml": TQuantityValueMilligramPerMilliliter,
+    "ng/ul": TQuantityValueNanogramPerMicroliter,
+    "ng/ml": TQuantityValueNanogramPerMilliliter,
+    "pg/ul": TQuantityValuePicogramPerMilliliter,
+}
+
+
+def _get_str(data_frame: pd.DataFrame, row: int, column: str) -> Optional[str]:
```
bdworth, line 56: I don't think we should allow this method to return `None`.

If `column` isn't found, we should throw.

If the value is na, we should return _some_ string. This will ensure that we have strings to operate on wherever this is called.

bdworth (changes requested):
not finished reviewing yet, but some initial comments

bdworth (changes requested):
couple more comments

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,272 @@
...
+                calculated_data_documents.append(self._get_260_230(data, i))
+
+        return calculated_data_documents
+
+    def _get_260_280(self, data: pd.DataFrame, row: int) -> CalculatedDataDocumentItem:
+
+        return CalculatedDataDocumentItem(
+            calculated_data_name="A260/280",
+            calculated_result=TQuantityValue(
+                value=_get_float(data, row, "260/280"), unit="unitless"
```
bdworth, line 138: I just merged #235, so please use its `UNITLESS` constant here and below.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bwolfe-benchling: I was mainly worried that inPlace modification of dataframes seems to be generally frowned upon: https://github.com/pandas-dev/pandas/issues/16529

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_reader.py
```diff
@@ -0,0 +1,25 @@
...
+class NanoDropEightReader:
+    @classmethod
+    def read(cls, contents: IOType) -> pd.DataFrame:
+        all_lines = lines_reader.read_to_lines(contents)
+        reader = CsvReader(all_lines)
+        lines = reader.pop_csv_block_as_lines()
+        raw_data = pd.read_csv(
+            StringIO("\n".join(lines)),
+            sep="\t",
+            dtype={"Plate ID": str, "Sample ID": str},
```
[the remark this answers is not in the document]
bwolfe-benchling: They are often numerical, but shouldn't be treated like numbers.  Pandas automatically converts to numeric types in those cases and I end up with added decimal precision on a filed that should be a string.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bwolfe-benchling: Some fields will accept None as a value and in that case will not be represented in the generated ASM.  Returning None simplifies the logic in those cases.  Would it make sense to have a method like _get_str that returns a atr or throws an error and a separate method like _get_str_or_none that acts like the current _get_str?

bdworth: Yeah, renaming this to `_get_str_or_none` and having `_get_str` return a `str` or throw sounds good. The latter should be used in all the places where you were calling `str` on an `Optional[str]`.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bdworth: ok, let's leave as is then. Thanks for future-proofing!

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,272 @@
...
+                        _get_float(data, row, a280_col)
+                    ),
+                )
+            )
+
+        return measurement_docs
+
+    def _get_concentration_col(self, data: pd.DataFrame) -> Optional[str]:
+        for col in data.columns:
+            if col.lower() in ["conc.", "conc", "concentration"]:
```
bdworth, line 270: missed this yesterday -- is this just a best guess based on files you've seen so far? Seems a tad hacky/hardcoded.

bdworth: It's probably ok to keep these, but if so let's extract a constant so this is more visible.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bdworth: I figured out how to do this. It does require some ugly union types though. I think I'll just push my changes.

bdworth: er, that's not a good operating model. So here's what you want:
1. Delete the declaration of `T` -- generics aren't helpful here.
2. Add this block above:
```
ConcentrationType = Union[
    TQuantityValueMicrogramPerMicroliter,
    TQuantityValueMicrogramPerMilliliter,
    TQuantityValueMilligramPerMilliliter,
    TQuantityValueNanogramPerMicroliter,
    TQuantityValueNanogramPerMilliliter,
    TQuantityValuePicogramPerMilliliter,
]
ConcentrationClassType = Union[
    type[TQuantityValueMicrogramPerMicroliter],
    type[TQuantityValueMicrogramPerMilliliter],
    type[TQuantityValueMilligramPerMilliliter],
    type[TQuantityValueNanogramPerMicroliter],
    type[TQuantityValueNanogramPerMilliliter],
    type[TQuantityValuePicogramPerMilliliter],
]
```
3. Add a type hint:
```
CONCENTRATION_UNIT_TO_TQUANTITY: Mapping[str, ConcentrationClassType] = {
``` 

4. This method should be:
```
def _get_concentration(
    conc: JsonFloat, unit: Optional[str]
) -> Optional[ConcentrationType]:
    if unit and unit in CONCENTRATION_UNIT_TO_TQUANTITY and isinstance(conc, float):
        cls = CONCENTRATION_UNIT_TO_TQUANTITY[unit]
        return cls(value=conc)

    return None
```

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bwolfe-benchling: yeah, this is mostly test driven since I don't have an official file spec.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bwolfe-benchling: for _get_str on a field that has None, do you have recommendation on error type and message?  Should this be AllotropeConversionError or something more general like ValueError?  As a message I was thinking something like "Error in _get_str:  Expecting str value but found None."

bwolfe-benchling: or maybe TypeError: _get_str does not support 'NoneType' values

bdworth: We're trying to return `AllotropeConversionError` for all exceptions thrown from allotropy. The most common pattern for null validation seems to be `assert_not_none`, so I'd just that.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bdworth: actually, better to use would be `TQuantityValueUnitless`

bdworth: P.S. It's fine to use either -- sorry for confusion.

nathan-stender: Actually - please use TQuantityValue with unit specified, as per the spec

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -0,0 +1,302 @@
...
+        return float(data_frame.iloc[row][column])
+    except (ValueError, TypeError):
+        return InvalidJsonFloat.NaN
+
+
+def _get_concentration(
+    conc: JsonFloat, unit: Optional[str]
+) -> Optional[ConcentrationType]:
+    if unit in CONCENTRATION_UNIT_TO_TQUANTITY and isinstance(conc, float):
+        cls = CONCENTRATION_UNIT_TO_TQUANTITY[str(unit)]
```
nathan-stender, line 104: nit: this `str(` call should not be necessary, if `unit` is in `CONCENTRATION_UNIT_TO_TQUANTITY`, the it must already be a string.

nathan-stender (approved):

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
bwolfe-benchling: I agree but the type checker is unhappy without the `str(` call.  Is there a better way?

nathan-stender: Oh of course it's the type checker 🙄 

I think if you add `if unit and unit in ...` to the front of the conditional it will be happy.

If that doesn't work, you can just leave it.

nathan-stender (approved):

PR #241, opened 2024-01-12
Make HasUnit units non-Optional

bdworth:
Depends on #228


nathan-stender (approved):

bdworth:
merged as part of #243

PR #245, opened 2024-01-12
feat: Add --regex argument to schema generation script

nathan-stender:
Adds a `--regex` argument, allowing generation of a subset of schemas, useful during testing of schema generation.


src/allotropy/allotrope/schema_parser/generate_schemas.py
```diff
@@ -81,6 +82,8 @@ def generate_schemas(root_dir: Path, *, dry_run: Optional[bool] = False) -> list
     for rel_schema_path in schema_paths:
         if str(rel_schema_path).startswith("shared"):
             continue
+        if schema_regex and not re.match(schema_regex, str(rel_schema_path)):
+            continue
```
bdworth, line 86: You could do something like this (changing L83-86), but also fine as written. One more use of `str(rel_schema_path)` and I'd ask for a local var.
```
        path_str = str(rel_schema_path)
        if path_str.startswith("shared") or (schema_regex and not re.match(schema_regex, path_str)):
            continue
```

bdworth (approved):
obviously you'll need to fix the formatting error, but otherwise lgtm. I'll re-approve once it's fixed.

src/allotropy/allotrope/schema_parser/generate_schemas.py
nathan-stender: You inspired me to use Pathlib for the shared check, removing a use of str(rel_schema_path)

I kept the conditions separate because it's a long line (lint would break up) and I think it keeps the logic clearer to have them separate.

bdworth (approved):

PR #246, opened 2024-01-12
feat: Add context manager to handle backups for schema generation script

nathan-stender:
This PR adds a context manager for backing up files during schema generation.

The main benefit of this is that it handles restoring a backups when an exception occurs, meaning that the files are not left in a messy state when there is an error during schema generation.

It also cleans up the code somewhat.


src/allotropy/allotrope/schema_parser/backup_manager.py
```diff
@@ -0,0 +1,62 @@
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+import shutil
+from typing import Optional, Union
+
+PathType = Union[Path, str]
+
+
+def files_equal(path1: PathType, path2: PathType) -> bool:
```
bdworth, line 10: nit: `files_equal`

src/allotropy/allotrope/schema_parser/backup_manager.py
```diff
@@ -0,0 +1,62 @@
...
+    with open(str(path1)) as file1, open(str(path2)) as file2:
+        for line1, line2 in zip(file1, file2):
```
bdworth, line 12: This won't fail if `file1` and `file2` have different numbers of lines but the first (number of lines in smaller file) lines pass your matching function.

The easiest way to fix would be add `strict=True` to `zip`, but that was introduced in 3.10 and we're still supporting 3.9. Seems like the best thing to do in 3.9 is use 
https://docs.python.org/3.3/library/itertools.html#itertools.zip_longest

PEP doc here: https://peps.python.org/pep-0618/

P.S. It appears this was copied from `generate_schemas` -- could you refactor? And since this is a refactor, I'm fine if you want to fix in follow-up.

src/allotropy/allotrope/schema_parser/backup_manager.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def _get_backup_path(path: PathType) -> Path:
+    _path = Path(path)
+    return Path(_path.parent, f".{_path.stem}.bak{_path.suffix}")
+
+
+def is_file_changed(path: PathType) -> bool:
+    if _get_backup_path(path).exists():
+        return not files_equal(path, _get_backup_path(path))
```
bdworth, line 25: nit: could extract `_get_backup_path(path)` to local var, here and in `restore_backup`

bdworth (changes requested):
found a bug in `files_equal` that could be fixed in follow up. Otherwise minor comments.

src/allotropy/allotrope/schema_parser/backup_manager.py
nathan-stender: Ah good catch - fixed. 

And whoops, I meant to delete `files_equal` from generate_schemas, done.

stephenworlow (approved):

bdworth (approved):

PR #247, opened 2024-01-12
fix: Rename TRelativeFluorescenceUnit and TRelativeLightUnit

nathan-stender:
These were mistakes in the manual editing of schemas, missing `QuantityValue`

These were caught by an upcoming change that will automatically generate all these references, removing the chance for this error in the future

But fix this case manually to reduce diffs in the larger PR


src/allotropy/allotrope/models/shared/definitions/custom.py
bdworth, line 1: These changes will conflict with my #243, but I'm fine fixing mine if this merges first.

bdworth (approved):
This will

PR #254, opened 2024-01-25
fix: rm duplicated calculated data related to group data in softmax

alejandro-salgado:
- remove duplicated calculated data documents related to group data in softmax


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -324,37 +345,42 @@ def _get_group_calc_docs(self, data: Data) -> list[CalculatedDataDocumentItem]:
         calculated_documents = []
         for group_block in data.block_list.group_blocks:
             for group_sample_data in group_block.group_data.sample_data:
+                for aggregated_entry in group_sample_data.aggregated_entries:
```
bdworth, line 348: could you please extract this new code to a helper? Triple-nested for loops are hard to follow.

Also, it seems like `data_sources` could be constructed outside this new loop since it uses `group_sample_data` but not `aggregated_entry`.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -324,37 +345,42 @@ def _get_group_calc_docs(self, data: Data) -> list[CalculatedDataDocumentItem]:
...
                         calculated_documents.append(
-                            CalculatedDataDocumentItem(
-                                calculated_data_identifier=random_uuid_str(),
-                                calculated_data_name=entry.name,
-                                calculation_description=group_block.group_columns.data.get(
-                                    entry.name
+                            self._build_calc_doc(
+                                name=entry.name,
+                                value=entry.value,
+                                data_sources=self._get_calc_docs_data_sources(
```
bdworth, line 376: again, can calculate data_sources outside the loop

bdworth (changes requested):
mostly ltgm, but a few suggestions for readabillity and (potentially) perf

bdworth (approved):
🚢 !

PR #261, opened 2024-01-30
fix: correct spectrophotometry schema

slopez-b:


src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
```diff
@@ -253,10 +266,31 @@ def _get_measurement_document(
...
+            if mass_concentration:
+                processed_data_aggregate_document = ProcessedDataAggregateDocument(
+                    processed_data_document=[
+                        ProcessedDataDocumentItem(
+                            # capture concentration on the A280 measurement document if the experiment type is
+                            # something other than DNA or RNA
+                            mass_concentration=mass_concentration
+                        )
+                    ]
+                )
```
bdworth, line 293: Out of scope for this PR, but could you please extract some helpers for this method? `_get_measurement_document` is now over 100 lines long. It was hard to tell, for instance, that these declarations
```
      na_type = _get_str_or_none(data, row, "NA Type")
      concentration_col = self._get_concentration_col(data)
```
needed to be outside the first `if` block.

tests/parsers/thermo_fisher_nanodrop_eight/testdata/Thermo_NanoDrop_Eight_example03.json
```diff
@@ -27,9 +27,15 @@
                                 "value": 0.185,
                                 "unit": "mAU"
                             },
+                            "processed data aggregate document": {
+                                "processed data document": [
+                                    {
```
bdworth, line 32: Did you add these lines manually? If they were generated I would expect to see the "mass conceentration" lines move right an intent or two. I think I'm ok with this for showing exactly what changed, but if we leave like this it might be worth following up to regenerate to make indenting look better

(and same question for all other changes to test JSON here)

bdworth (approved):
two non-blocking questions -- feel free to resolve so you can merge.

src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_parser.py
slopez-b: @bdworth I was tempted to make those changes but then I realized I'd end up rewriting a good part of this parser. This PR is supposed to only make changes to the schema.

There is actually an issue with the way we are parsing the `mass_concentration` that is outside the scope of this PR and I plan to handle after discussing with Joe.

tests/parsers/thermo_fisher_nanodrop_eight/testdata/Thermo_NanoDrop_Eight_example03.json
slopez-b: @bdworth I did not add the lines manually. I generated the files with the parser, but unstaged the identifier lines so it was easier to see what had changed on the files. It seems like doing that broke the formatting somehow. I'll fix it.

bdworth (approved):

PR #263, opened 2024-01-31
fix: make calculated data documents optional for softmax pro

keithmacaulay:


src/allotropy/allotrope/models/plate_reader_benchling_2023_09_plate_reader.py
```diff
@@ -359,7 +359,7 @@ class CalculatedDataDocumentItem:
 
 @dataclass
 class CalculatedDataAggregateDocument:
-    calculated_data_document: list[CalculatedDataDocumentItem]
+    calculated_data_document: Optional[list[CalculatedDataDocumentItem]] = None
```
slopez-b, line 362: `calculated data document`(s) should not be optional. If there is a `calculated data aggregate document` (which is optional), it should have at least one `calculated data document`.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -97,7 +97,11 @@ def _get_model(self, file_name: str, data: Data) -> Model:
                     for position in plate_block.iter_wells()
                 ],
                 calculated_data_aggregate_document=CalculatedDataAggregateDocument(
-                    calculated_data_document=self._get_calc_docs(data)
+                    calculated_data_document=(
+                        None
```
slopez-b, line 101: What you can do here instead is check if `_get_calc_docs` returns any items and based on that include or not the `calculated_data_aggregate_document`, which is optional, to avoid having an "aggregate doc" without any "document".

slopez-b (changes requested):

nathan-stender:
@alejandro-salgado took care of here: https://github.com/Benchling-Open-Source/allotropy/pull/264

keithmacaulay:
Closing as resolved here: #264

PR #264, opened 2024-01-31
fix: Do not add empty list of calculated data documents to Moldev Softmax Pro output

alejandro-salgado:
- Do not add empty list of calculated data documents to Moldev Softmax Pro output


nathan-stender (approved):

nathan-stender (changes requested):
LGTM - please add changelog entry

nathan-stender (approved):

PR #268, opened 2024-01-31
Fix: Vi-cell xr parser invalid row and required column

slopez-b:
- Update vi-cell XR parser so that `Total cells` column is no longer required, (as per the latest version of the schema).
- Ignore invalid first row when present for vi-cell XR files.


src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_parser.py
```diff
@@ -144,7 +143,9 @@ def _get_cell_counting_document_item(
                                     ),
                                     viability__cell_counter_=viability__cell_counter_,
                                     viable_cell_density__cell_counter_=viable_cell_density__cell_counter_,
-                                    total_cell_count=total_cell_count,
+                                    total_cell_count=get_property_from_sample(
+                                        sample, "Total cells"
```
bdworth, line 147: Could you create a constant for "Total cells" and use that here and in `property_lookup`?

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
```diff
@@ -28,13 +28,19 @@ def _read_data(self) -> pd.DataFrame:
...
         skiprows = header_row + 1
+        date_header = DATE_HEADER[self.file_version]
+
         file_data: pd.DataFrame = pd.read_excel(
-            self.contents,
-            skiprows=skiprows,
-            names=header,
-            parse_dates=[DATE_HEADER[self.file_version]],
-            date_format="%d %b %Y  %I:%M:%S %p",
+            self.contents, skiprows=skiprows, names=header
```
bdworth, line 34: fine, but just noting this will conflict with #262 if it merges first

tests/parsers/beckman_vi_cell_xr/testdata/v2.04/Beckman_Vi-Cell-XR_example02_instrumentOutput.xls
bdworth, line 1: Was the file in its former state not a valid instrument output file?

If the former file was a valid instrument output file, should we leave it in and add a new file rather than replacing?

bdworth (changes requested):
mostly lgtm, but a couple comments

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_parser.py
slopez-b: I'll do it for all properties in that lookup dict

bdworth: better idea: #269

tests/parsers/beckman_vi_cell_xr/testdata/v2.04/Beckman_Vi-Cell-XR_example02_instrumentOutput.xls
slopez-b: This was an invalid instrument output file that was used to test that if `Total cells` was not present the parser would raise an `AllotropeConversionError`. 

After the changes in this PR (`Total cells` being now optional), the test failed as expected, so I changed the file to remove a different column that is still required and keep the test with the same file.

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_parser.py
slopez-b: @bdworth I think is a good idea but also think this should be blocking. Can we merge this PR and then proceed with the other changes?

bdworth: #269 has merged. Sorry, I only saw this comment now.

bdworth: I should be able to approve quickly once you fix conflicts!

bdworth (changes requested):
There are merge conflicts to fix. (Sorry!)

PR #270, opened 2024-02-01
Move functions into _Sample class

bdworth:
OOP++


src/allotropy/parsers/beckman_vi_cell_blu/vi_cell_blu_parser.py
```diff
@@ -62,33 +63,34 @@ def __init__(self, column_name: str, data_type: Any) -> None:
...
-    if value is None:
-        msg = f"Missing value for column '{column}'."
-        raise AllotropeConversionError(msg)
-    return value
+    def get_value(self, column: str) -> Optional[Any]:
+        if column not in self.data_frame.columns:
+            return None
+        return self.data_frame[column][self.row]
+
+    def get_value_not_none(self, column: str) -> Any:
```
rayaankha-tara, line 76: Any reason for this to not be a boolean flag on `get_value`. I dont have a strong feeling just curious as to your thoughts

rayaankha-tara (approved):

src/allotropy/parsers/beckman_vi_cell_blu/vi_cell_blu_parser.py
bdworth: It could be, but I try to limit my refactors to doing one thing at a time. Moving functions to methods and changing their signatures is two things :)

PR #271, opened 2024-02-01
feat: Replace null with N/A as empty value in Moldev Softmax Pro adapter

alejandro-salgado:
- Replace null with N/A as empty value in Moldev Softmax Pro adapter


nathan-stender (changes requested):
Please just replace "null" with "N/A" in these test files, instead of changing all the IDs

We will fix this someday I promise 😭

alejandro-salgado:
done in https://github.com/Benchling-Open-Source/allotropy/pull/272

PR #272, opened 2024-02-01
feat: Replace null with N/A in Moldev Softmax Pro

alejandro-salgado:
-  Replace null with N/A in Moldev Softmax Pro


alejandro-salgado:
So much clean this way, thanks for the advice @nathan-stender .

nathan-stender (approved):

PR #277, opened 2024-02-08
fix: Fix swapped fields in Unchained Labs Lunatic adapter

nathan-stender:


rayaankha-tara (approved):
Why did the order of fields in the `sample document` property in the test json need to swap? Is it auto generated?

nathan-stender:
@rayaankha-tara no, it was easier to change the field names than to change the field values in every case.

PR #284, opened 2024-02-14
Inline test_utils.model_from_file()

bdworth:
There doesn't seem to be a good reason to wrap this for tests with only one test caller.


nathan-stender (commented):
Good cleanup. I believe the test function predates the util function

nathan-stender (dismissed):

nathan-stender (approved):

PR #289, opened 2024-02-14
fix: Handle invalid well values in SoftmaxPro

nathan-stender:
Populate invalid values (e.g. "Path?" from provided example) with "NaN"


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -291,7 +292,7 @@ class DataElement:
     temperature: Optional[float]
     wavelength: float
     position: str
-    value: float
+    value: float | InvalidJsonFloat
```
bdworth, line 295: `JsonFloat` is an alias for `float | InvalidJsonFloat`

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -331,10 +332,7 @@ def create(
                     temperature=temperature,
                     wavelength=wavelength,
                     position=str(position),
-                    value=try_float(
-                        value,
-                        f"value from block plate {plate_name} and wavelength {wavelength}",
-                    ),
+                    value=try_float_or_none(value) or InvalidJsonFloat.NaN,
```
bdworth, line 335: This is fine, but anticipating that we might want to do more of this, how about adding a utility method a la the following and calling it here?
```
def try_float_or_nan(value: Optional[str]) -> JsonFloat:
    if not value:
        return InvalidJsonFloat.NaN
    try:
        return float(value)
    except ValueError:
        return InvalidJsonFloat.NaN
```

bdworth (approved):
fine as is, but two suggestions

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
nathan-stender: There are some other spots that use this already, plan to follow up with a util.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
nathan-stender: Will get this in util PR as well

PR #292, opened 2024-02-15
Add to InvalidJsonFloat docstring

bdworth:
Explain enum names and (current) lack of use of the last two enum values.


nathan-stender (changes requested):
These were auto-generated by schema generation code, so we are going to stick with them. Please don't remove.

nathan-stender (approved):

PR #295, opened 2024-02-15
Refactor random UUID generation (take 1)

bdworth:
These changes will facilitate mocking `VendorParser._get_id_generator()` in tests to create stable identifiers. Not to mention potentially allowing changing id generation in src).

Note that making the id generator an instance method on `VendorParser` rather than a singleton accessed by a function makes mocking in tests significantly easier. If this were accessed via a function, we would have to mock each import differently in tests.

Steps taken:
1. Move `random_uuid_str` to vendor_parser module
2. Create classmethod of same name in VendorParser that calls function and use it in children
3. Create IdGenerator interface and supply a default implementation
4. Delete utility function
5. Create utility function in each module that was still importing it, using the child of VendorParser appropriate to the module

### DIFF
--- a/src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
+++ b/src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
@@ -22,7 +22,6 @@ from allotropy.allotrope.pandas_util import read_csv
 from allotropy.parsers.appbio_quantstudio.calculated_document import CalculatedDocument
 from allotropy.parsers.appbio_quantstudio.referenceable import Referenceable
 from allotropy.parsers.lines_reader import LinesReader
-from allotropy.parsers.moldev_softmax_pro.softmax_pro_parser import SoftmaxproParser
 from allotropy.parsers.utils.values import (
     assert_not_empty_df,
     assert_not_none,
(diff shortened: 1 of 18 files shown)


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -39,6 +38,12 @@
 )
 
 
+def random_uuid_str() -> str:
+    from allotropy.parsers.moldev_softmax_pro.softmax_pro_parser import SoftmaxproParser
+
+    return SoftmaxproParser.random_uuid_str()
```
slopez-b, line 44: @bdworth should this function actually depend on `SoftpaxproParser` implementation of `random_uuid_str`?

What if the parser is deprecated in the future? I personally think parsers should be independent from each other.

slopez-b: also applies for `src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py`

slopez-b (changes requested):

bdworth:
replaced by #309

PR #296, opened 2024-02-15
Generate test JSON files with stable identifiers (take 1)

bdworth:
Depends on #295
Includes #299
(both should merge before this does)

Test-only changes (after #299 merges):
1. Generate stable ids in tests.
2. Test all identifier fields in tests; previously some were excluded (and the mechanism for doing that will be removed in follow-up).
3. Add the ability to write to existing files if the actual contents don't match the expected ones.
4. Regenerate output json files for all failing tests by temporarily setting `write_actual_to_expected_on_fail` to `True` in `test_utils.validate_contents` (and not committing that change).

Keep changes to existing test JSON files as small as possible by preserving indent level (some are 2, and some are 4).


src/allotropy/parsers/beckman_vi_cell_blu/vi_cell_blu_parser.py
```diff
@@ -70,7 +71,13 @@ class _Sample:
     def get_value(self, column: str) -> Optional[Any]:
         if column not in self.data_frame.columns:
             return None
-        return self.data_frame[column][self.row]
+        value = self.data_frame[column][self.row]
```
bdworth, line 74: This change should probably go in its own PR.

tests/parsers/beckman_vi_cell_xr/testdata/v2.06/Beckman_Vi-Cell-XR_hiddenRow.json
```diff
@@ -43,7 +43,7 @@
                                         },
                                         "average total cell diameter": {
                                             "value": 16.5,
-                                            "unit": "μm"
+                                            "unit": "\u03bcm"
```
bdworth, line 46: todo: fix this

tests/parsers/beckman_vi_cell_xr/testdata/v2.06/Beckman_Vi-Cell-XR_hiddenRow.json
```diff
@@ -70,7 +70,7 @@
                     "measurement document": [
                         {
                             "measurement time": "2023-04-06T10:28:50+00:00",
-                            "measurement identifier": "d1e18e47-9795-4ead-81ed-f1cb74832be0",
+                            "measurement identifier": "BECKMAN_VI_CELL_XR_TEST_ID_0",
```
bdworth, line 73: sigh, id numbers are always 0

bdworth (commented):
things to follow up on

src/allotropy/parsers/vendor_parser.py
```diff
@@ -10,6 +10,17 @@
 from allotropy.parsers.utils.values import assert_not_none
 
 
+class IdGenerator(ABC):
```
bdworth, line 13: This should probably be moved up one PR as well.

PR #306, opened 2024-02-19
Pin mypy to 1.7.1

bdworth:


bdworth:
fixed the issue here in #308

PR #307, opened 2024-02-19
Pin mypy to 1.6.0

bdworth:
Current version is 1.8.0, but for some reason hatch was apparently using 1.6.0 until just recently. Now it's using a later version, and no changes can be committed without fixing or suppressing some warnings.

Pin to 1.6.0 to unblock development, and follow up later to upgrade mypy.


bdworth:
fixed the issue here in #308

PR #310, opened 2024-02-19
Generate stable ids in tests

bdworth:
Make ids in test output files both stable and unique within their file. The exact generation can be tweaked later as desired; the important part is making ids stable so id lines don't churn every time a src change makes it necessary to regenerate a test output file.

1. Add stable id generation mechanism via mock.patch.
2. Test all identifier fields in tests; previously some were excluded (and the mechanism for doing that will be removed in follow-up).
3. Add the ability to write to existing files if the actual contents don't match the expected ones.
4. Regenerate output json files for all failing tests by temporarily setting `write_actual_to_expected_on_fail` to `True` in `test_utils.validate_contents` (and not committing that change).

Keep changes to existing test JSON files as small as possible by preserving indent level (some are 2, and some are 4). Other changes that will be observed in changes to output JSON files:
1. Adding newlines to ends of some files
2. "ASM converter version" values changing: unavoidable as this is part of the JSON output (and already ignored in tests).
3. dictionary order changing
4. single lines split out into multiple lines
5. UTF-8 encoding improvements -- e.g., "\u03bcm" to "μm"


nathan-stender (dismissed):

nathan-stender (approved):

bdworth:
@nathan-stender fixed semantic merge conflict with 4bab77b4bf3f3947125e67ea244d0bbafc88925e

PR #316, opened 2024-02-19
fix: Update the Wayland-Yutani tutorial to use hatch for running the script

nathan-stender:
Also mention the ability to update test files with the test function

### DIFF
--- a/tests/parsers/example_weyland_yutani/testdata/Weyland_Yutani_simple_correct.json
+++ b/tests/parsers/example_weyland_yutani/testdata/Weyland_Yutani_simple_correct.json
@@ -1,6 +1,6 @@
 {
     "measurement aggregate document": {
-        "measurement identifier": "3429be41-a354-4ea3-b2cd-73eb0640e908",
+        "measurement identifier": "EXAMPLE_WEYLAND_YUTANI_TEST_ID_0",
         "plate well count": {
             "value": 16,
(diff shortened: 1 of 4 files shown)


dchou-benchling (commented):
Doesn't look like I have powers to approve but thank you for doing this!
Also noting that there are several other scripts in that folder that we may want to do this for in the future but this is the only one called out in the tutorial

tests/parsers/example_weyland_yutani/testdata/Weyland_Yutani_simple_correct.json
```diff
@@ -1,6 +1,6 @@
 {
     "measurement aggregate document": {
-        "measurement identifier": "EXAMPLE_WEYLAND_YUTANI_TEST_ID_0",
+        "measurement identifier": "3429be41-a354-4ea3-b2cd-73eb0640e908",
```
mdlu, line 3: could you put back the old ID?

mdlu (changes requested):

tests/parsers/example_weyland_yutani/testdata/Weyland_Yutani_simple_correct.json
nathan-stender: Done tyty

mdlu (approved):

PR #324, opened 2024-02-28
Chore: Adding SMP test data - examples of exports containing partial plates

LimulusPolyphemus:


nathan-stender (changes requested):
Thanks for adding the test data!

That said, can you please add something to these names indicating that their test case, e.g. `MD_SMP_absorbance_endpoint_partial_plate_example01.txt` ?

We've been lax with this, but it would be good to know what test cases are testing for future reference.

LimulusPolyphemus:
@nathan-stender I updated the file names to reflect the test case.

nathan-stender:
@LimulusPolyphemus I think you need to push the commit

LimulusPolyphemus:
@nathan-stender I did. The commit is on the branch, but it's not appearing on the PR for some reason, hmmm

nathan-stender (approved):

nathan-stender:
The change appeared 👍

PR #325, opened 2024-03-01
Move VendorType to to_allotrope

bdworth:
Use VendorType (Union of str and Vendor enum) in to_allotrope but not outside that. parser_factory shouldn't be part of our most public functions.

Out of scope: moving Vendor out of `parser_factory`. (It seems odd for Vendor to imported from `parser_factory`.)


nathan-stender (changes requested):
I don't think this is something we should insist on. It will make working with, e.g. config files for setting up tooling harder. We already handle the error if the string is wrong right at the front.

nathan-stender (approved):

PR #327, opened 2024-03-05
Replace float nan with NaN string in Moldev Softmax Pro

alejandro-salgado:


src/allotropy/parsers/utils/values.py
```diff
@@ -46,7 +60,7 @@ def try_float_or_none(value: Optional[str]) -> Optional[float]:
 
 
 def try_float_or_nan(value: Optional[str]) -> JsonFloat:
-    float_value = try_float_or_none(value)
+    float_value = try_non_nan_float_or_none(value)
```
slopez-b, line 63: @alejandro-salgado is this really necesary? i.e. do we get NaN values from any of the instrument files?

My understanding is that it is either a float or `None`

alejandro-salgado: this is a safe guard in case there is a value that is read as nan.

slopez-b: I don't really think this is necessary `try_float_or_none` should be enough as in all the other cases. It is very unlikely we would end up with something that is going to be serialized as `NaN` by the parsers unless we explicitly pass `np.nan` or `float(nan)`

src/allotropy/parsers/utils/values.py
```diff
@@ -38,6 +39,19 @@ def try_float(value: str, value_name: str) -> float:
         raise AllotropeConversionError(msg) from e
 
 
+def try_non_nan_float(value: str) -> float:
```
slopez-b, line 42: @alejandro-salgado is this util function used anywhere? I only see the definition.

src/allotropy/parsers/utils/values.py
alejandro-salgado: The idea of this function to be used anywhere, if we do not check for nan value then the rest of the logic will not produce the correct result. If we read a float(nan) value with try_float_or_none the result will be float(nan) and since that is not None the result of the function will be float(nan) and it should be "NaN"

src/allotropy/parsers/utils/values.py
alejandro-salgado: It is an alternative to try_float in which an error is raised if a nan value is read.

slopez-b (approved):

PR #328, opened 2024-03-05
chore: Add CHANGE_NOTES detailing differences between allotropy schemas and published ASM

nathan-stender:
Add CHANGE_NOTES for each schema that has changes from published ASM

The notes seek to explain what is different, why it was changed, and what the proposed action is, either to the schema, parsers, or to future versions of ASM.

Much of the explanation/proposal sections are currently TODO as we backfill existing changes, but part of the goal here is to establish a pattern so that going forward these will be filled out as changes are made.


src/allotropy/allotrope/schemas/pcr/BENCHLING/2023/09/CHANGE_NOTES.md
LimulusPolyphemus, line 1: @nathan-stender for "data system document", see comment above.

For the "device document" within the "device system document" and the additional fields within the "device control document", it probably was not necessary to remove these - as they are just unused.

Agree with "diagnostic trace", "processed data" and "statistics document", they just weren't added because they were not being used.

Removed "written name", "vial location identifier", and "location identifier" from "sample document". These should not have been removed -especially since we've been normalizing on using "location identifier".

The "sample role type" should be changed back to an enum for consistency.

The "reference DNA description" and "reference sample description" were added to the "data processing document" within the "calculated document" to assist in the interpretation of the "calculated result" the core schema supports the use of a "data processing document" within the "calculated data document", however, the constituent fields are not declared.

The "target DNA description" and "PCR detection chemistry" were proposed as required, but the terms were not instantiated as required when the model was updated.

Similarly, the "processed data document" within the "measurement document" was proposed as required, but was that was not accepted - which upon reflection is probably correct.

Correct, the "data source identifier" and "data source feature" should be conditionally required if the "data source document" is used.

It has been requested to relax the requirement for "analyst" since it is not always provided.
https://gitlab.com/allotrope/adm/-/issues/615

Agree with the name changes which derived from splitting the models, but it should be "PCR" and not "PRC"

All of the "time" terms were originally proposed as such, but they were changed to "duration" when the terms were created because "duration" was more semantically correct.

Similarly, "primer" was dropped from those terms, so that the terms were not overly context specific.

src/allotropy/allotrope/schemas/plate-reader/BENCHLING/2023/09/CHANGE_NOTES.md
LimulusPolyphemus, line 1: @nathan-stender "asset management identifier" was never meant to be a required field. That was a bug when Allotrope published the initial schema.

"compartment temperate" was relaxed because it was not universally provided by all instruments.

See note above regarding "data system document"

src/allotropy/allotrope/schemas/ultraviolet-absorbance/BENCHLING/2023/09/CHANGE_NOTES.md
LimulusPolyphemus, line 1: The comments here are the same as those for the "fluorescence" and "luminescence" measurements.

LimulusPolyphemus (commented):
@nathan-stender I provided comments on all of those schema differences.

src/allotropy/allotrope/schemas/cell-culture-analyzer/BENCHLING/2023/09/CHANGE_NOTES.md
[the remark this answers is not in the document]
nathan-stender: Noted, thank you

src/allotropy/allotrope/schemas/fluorescence/BENCHLING/2023/09/CHANGE_NOTES.md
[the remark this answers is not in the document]
nathan-stender: I think we should remove it once current parsers are migrated off of it, but agilent_gen5 is still using it at this time.

nathan-stender (commented):
Thanks Joe

src/allotropy/allotrope/schemas/ultraviolet-absorbance/BENCHLING/2023/09/CHANGE_NOTES.md
```diff
@@ -0,0 +1,21 @@
+DEPRECATED - TO NOT USE. This is replaced by plate-reader/BENCHLING/2023/09/plate-reader.json
```
stephenworlow, line 1:
```suggestion
DEPRECATED - DO NOT USE. This is replaced by plate-reader/BENCHLING/2023/09/plate-reader.json
```

src/allotropy/allotrope/schemas/pcr/BENCHLING/2023/09/CHANGE_NOTES.md
```diff
@@ -0,0 +1,72 @@
...
+* Removed "device document" from "device system document"
+  * Reasoning: contents of "device document" are covered in other documents
+    * All fields are covered between "device system document" and "device control document"
+  * Proposal: TODO these should be added back to conform with ASM since they are optional.
+
+* Removed "product manufacturer", "brand name", "equipment serial number", "model number", "firmware version" from "device control document" in favor of "device system document"
+  * Reasoning: these fields are covered in "device system document"
+  * Proposal: TODO these should be added back to conform with ASM since they are optional.
+
+* Missing "processed data aggregate document" and "statistics aggregate document" from technique aggreagete document
```
stephenworlow, line 18:
```suggestion
* Missing "processed data aggregate document" and "statistics aggregate document" from technique aggregate document
```

stephenworlow (approved):

PR #333, opened 2024-03-06
fix: Update SoftmaxPro parser to iterate over all wells of a plate

bwolfe-benchling:
Update of Softmax Pro structure to iterate all wells of a plate regardless of the number of populated rows or columns.  Rather than relying on the Plate Block header values for first column, number of columns, first row, and number of rows, use the number of wells to determine the plate dimensions and process all positions.


bdworth (dismissed):
mostly lgtm -- just one comment

nathan-stender (approved):
LGTM - thanks for all the test additions!

nathan-stender (approved):

PR #336, opened 2024-03-07
Add encoding to NamedFileContents

bdworth:
Add encoding to NamedFileContents and use it when decoding (bytes) further down.

Keep existing behavior using chardet with no encoding, but will be removed shortly.

Note that there is a functional change in `read_to_lines`, which (given no changes to the call stack) will no longer assume UTF-8 but instead use chardet. This will also change shortly.


tests/parsers/appbio_quantstudio/appbio_quantstudio_structure_test.py
```diff
@@ -142,7 +142,7 @@ def test_header_builder_invalid_plate_well_count() -> None:
 @pytest.mark.short
 def test_header_builder_no_header_then_raise() -> None:
     header_contents = get_raw_header_contents(raw_text="")
-    lines = _read_to_lines(header_contents)
+    lines = _read_to_lines(header_contents, encoding="UTF-8")
```
bdworth, line 145: without these changes, we get `allotropy.exceptions.AllotropeConversionError: Unable to detect text encoding for file. The file may be empty.`

tests/parsers/agilent_gen5/to_allotrope_test.py
```diff
@@ -48,7 +50,8 @@ def test_to_allotrope_absorbance(filename: str) -> None:
     expected_filepath = (
         f"tests/parsers/agilent_gen5/testdata/absorbance/{filename}.json"
     )
-    allotrope_dict = from_file(test_filepath, VENDOR_TYPE)
+    encoding = "UTF-16" if filename == UTF_16_FILENAME else None
```
nathan-stender, line 53: It seems from this like this change is somehow enforcing encoding somewhere where it was previously being auto-detected? 

That makes me a bit worried about pushing this through. This seems like the kind of change that should have no test changes.

Can you tell why this is needing the encoding passed now?

nathan-stender (changes requested):

tests/parsers/agilent_gen5/to_allotrope_test.py
bdworth: it doesn't -- this was me wanting to explicitly test passing an encoding in. This test passes if I revert all the changes here, and I agree that it's more reassuring to make these changes with no changes here. So I'll revert shortly.

src/allotropy/parsers/lines_reader.py
```diff
@@ -13,10 +13,9 @@
 EMPTY_STR_PATTERN = r"^\s*$"
 
 
-def read_to_lines(
-    named_file_contents: NamedFileContents, encoding: Optional[str] = "UTF-8"
```
bdworth, line 17: called out in the PR description that this is a functional change

nathan-stender (dismissed):

nathan-stender (approved):

PR #349, opened 2024-03-11
Add optical imaging to plate reader schema

alejandro-salgado:
- Add optical imaging to plate reader schema


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -143,13 +145,7 @@ def _get_plate_reader_document_item(
...
     def _get_fluorescence_measurement_document(
         self, plate_block: PlateBlock, position: str
-    ) -> list[
-        Union[
-            UltravioletAbsorbancePointDetectionMeasurementDocumentItems,
-            FluorescencePointDetectionMeasurementDocumentItems,
-            LuminescencePointDetectionMeasurementDocumentItems,
-        ]
-    ]:
+    ) -> list[MeasurementDocumentItems]:
```
slopez-b, line 148: This and all the other `_get_<measurement_document_type>_measurement_document` methods can only return one type of measurement, the signature should be consistent with that.

```suggestion
    ) -> list[FluorescencePointDetectionMeasurementDocumentItems]:
```

slopez-b (changes requested):
Looks really good. Just a question abot some method signatures.

src/allotropy/allotrope/schemas/plate-reader/BENCHLING/2023/09/plate-reader.json
```diff
@@ -474,6 +480,53 @@
...
+                    ],
+                    "properties": {
+                      "image feature document": {
+                        "type": "array",
+                        "$asm.array-ordered": true,
+                        "$asm.array-mixed": false,
+                        "$asm.pattern": "indexed datum",
+                        "items": {
+                          "type": "object",
+                          "$asm.property-class": "http://purl.allotrope.org/ontologies/result#AFR_0002659",
```
james-leinas, line 499: @alejandro-salgado, image feature document will be a new term to AFO so property class should be "TODO".

james-leinas (commented):
@alejandro-salgado, looks great! Just a couple items about the ASM property classes but structure and fields match original proposal. Thanks!

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
alejandro-salgado: The definition of measurement_document that is assigned with the result of this functions is MeasurementDocumentItems, so I can't define the return type as FluorescencePointDetectionMeasurementDocumentItems because I get 

**expression has type "list[FluorescencePointDetectionMeasurementDocumentItems]", variable has type "MeasurementDocumentItems"**

nathan-stender: @alejandro-salgado I pushed a change to the PR to fix this. By changing the typing of `measurement_document` to `Sequence` during dynamic assignment, and then casting it to a list, you can make the function signatures specific.

slopez-b (approved):

nathan-stender (approved):

PR #354, opened 2024-03-15
refactor: remove unsupported files and read modes in gen5 adapter

slopez-b:
First part of the new implementation of Gen5 adapter

- Remove support for kinetic measurements in Gen5 adapter
- Remove support for multiplate files in Gen5 adapter
- Remove unused Gen5 structure classes and stop parsing sections that are not used in the adapter
- Parse more metadata from Procedure Details Section


src/allotropy/allotrope/models/plate_reader_benchling_2023_09_plate_reader.py
```diff
@@ -1,6 +1,6 @@
 # generated by datamodel-codegen:
 #   filename:  plate-reader.json
-#   timestamp: 2024-03-11T14:24:12+00:00
+#   timestamp: 2024-03-15T12:07:47+00:00
```
nathan-stender, line 3: Can you undo this change, as there do not seem to be any changes in the schema. (there probably were at one point, and then they were reverted, which caused the generation script to update the timestamp, if you revert it it should not change again)

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -3,23 +3,24 @@
...
-    plates: list[PlateData]
+    plate_data: PlateData
 
     @staticmethod
     def create(section_lines_reader: SectionLinesReader) -> Data:
-        plates: list[PlateData] = [
-            PlateData.create(lines_reader)
-            for lines_reader in section_lines_reader.iter_sections("^Software Version")
-        ]
+        plate_readers = list(section_lines_reader.iter_sections("^Software Version"))
```
nathan-stender, line 17: Can you change this back to `plates` or perhaps `plate_sections`? 

`plate_readers` makes it sound like these are reader classes, which could be confusing.

nathan-stender (changes requested):

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
slopez-b: I changed it to plate_readers because `SectionLinesReader.iter_sections` yields `LinesReader` instances. But I agree it may be not the best wording for it.

nathan-stender (approved):

PR #355, opened 2024-03-19
Chore: Adding additional Lunatic test file

LimulusPolyphemus:
Adding test file for Unchained Labs Lunatic for an dsDNA measurement where the wavelength is not present within the "Concentration" header.


nathan-stender (dismissed):
Is there a reason the filename has `concentrationHeader` instead of `concentration_header`?

Not gonna block as it's not that big of a deal, but seems odd?

nathan-stender (approved):

PR #356, opened 2024-04-01
Add Revvity Kaleido adapter

alejandro-salgado:
- Add Revvity Kaleido adapter and related tests.


src/allotropy/parsers/revvity_kaleido/kaleido_common_structure.py
```diff
@@ -0,0 +1,56 @@
...
+    "E": SampleRoleType.control_sample_role.value,
+    "ZL": SampleRoleType.control_sample_role.value,
+    "ZH": SampleRoleType.control_sample_role.value,
+    "LB": SampleRoleType.control_sample_role.value,
+    "LC": SampleRoleType.control_sample_role.value,
+    "LH": SampleRoleType.control_sample_role.value,
+}
+
+
+SCAN_POSITION_CONVERTION = {
```
nathan-stender, line 37:
```suggestion
SCAN_POSITION_CONVERSION = {
```

src/allotropy/parsers/revvity_kaleido/kaleido_common_structure.py
```diff
@@ -0,0 +1,56 @@
...
+    "TOP": ScanPositionSettingPlateReader.top_scan_position__plate_reader_,
+    "BOTTOM": ScanPositionSettingPlateReader.bottom_scan_position__plate_reader_,
+}
+
+
+TRANSMITTED_LIGHT_CONVERTION = {
```
nathan-stender, line 43:
```suggestion
TRANSMITTED_LIGHT_CONVERSION = {
```

src/allotropy/parsers/revvity_kaleido/kaleido_structure_v3.py
```diff
@@ -0,0 +1,177 @@
...
+        if str(column).startswith("Unnamed"):
+            results = results.drop(columns=column)
+
+    return Results(
+        barcode=barcode,
+        results=results,
+    )
+
+
+def create_analysis_results(reader: CsvReader) -> list[AnalysisResult]:
```
nathan-stender, line 65: It looks to me like this is identical to v2, move to common?

src/allotropy/parsers/revvity_kaleido/kaleido_common_structure.py
```diff
@@ -0,0 +1,56 @@
...
+}
+
+
+@dataclass(frozen=True)
+class WellPosition:
+    column: str
+    row: str
+
+    def __repr__(self) -> str:
+        return self.row + self.column
```
nathan-stender, line 56: Is this used anywhere, if used, seems like there should be some separator, like a "," or ".", if not used, remove.

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,410 @@
...
+
+@dataclass
+class BackgroundInfo:
+    experiment_type: str
+
+
+@dataclass(frozen=True)
+class Results:
+    barcode: str
+    results: pd.DataFrame
```
nathan-stender, line 38: Storing the raw dataframe in the this dataclass is an anti-pattern.

You should instead process the data in a create function and store it in the dataclass.

The reason for this is that it makes reasoning about the structure and testing easier.

In this example, you should could, for example, create a two-dim array of wells.

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,410 @@
...
+            return str(self.results.loc[well_position.row, well_position.column])
+        except KeyError as e:
+            error = f"Unable to get well at position '{well_position}' from results section."
+            raise AllotropeConversionError(error) from e
+
+
+@dataclass(frozen=True)
+class AnalysisResult:
+    analysis_parameter: str
+    results: pd.DataFrame
```
nathan-stender, line 69: Same comment as above.

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,410 @@
...
+            reader.drop_until_inclusive("^Plate Type"),
+            msg="Unable to find Plate Type section.",
+        )
+        reader.drop_until("^Platemap")
+        return PlateType({})
+
+
+@dataclass(frozen=True)
+class Platemap:
+    data: pd.DataFrame
```
nathan-stender, line 182: Same comment as above. For this one, see perkin elmer envision plate map for an example.

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,410 @@
...
+    version: str
+    background_info: BackgroundInfo
+    results: Results
+    analysis_results: list[AnalysisResult]
+    measurement_info: MeasurementInfo
+    plate_type: PlateType
+    platemap: Platemap
+    measurements: Measurements
+
+    def get_experiment_type(self) -> str:
```
nathan-stender, line 355: Generally, you should not need/want to create accessor classes for nested dataclasses, you can just access them directly. 

e.g. replace `data.get_experiment_type()` with `data.background_info.experiment_type`

The dataclass names/structure should be self documenting and simple access, removing the need to these functions.

src/allotropy/parsers/revvity_kaleido/kaleido_structure_v3.py
alejandro-salgado: There is a difference in the section titles, one is Measurement Basic Information, the other is Measurement Information.

src/allotropy/parsers/revvity_kaleido/kaleido_common_structure.py
alejandro-salgado: It is used in error messages. The idea is to show for example "A1" or "H12", that is why there is no separator.

alejandro-salgado: Now that there are no dataframes in the structure this dataclass is no longer needed.

src/allotropy/parsers/revvity_kaleido/kaleido_structure_v3.py
nathan-stender: Ah, I did not spot that, thanks 👍

nathan-stender (changes requested):
There are still a lot of run-time lookup / unnecessary accessors in these classes.

At creation time, anything that can be stored should be. You should create descriptive, named attributes in the dataclass and access them directly.

slopez-b (approved):

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,369 @@
...
+            logging.warning(
+                f"Unable to read analysis result '{analysis_parameter}'. Ignoring"
+            )
+            return None
+
+        for column in results:
+            if str(column).startswith("Unnamed"):
+                results = results.drop(columns=column)
+
+        try:
```
nathan-stender, line 131: Can you add a comment explaining what this is going, specifically, the `try_float_or_none` check.

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,435 @@
...
+        return AnalysisResult(
+            analysis_parameter=analysis_parameter,
+            results={
+                f"{row}{col}": values[col]
+                for row, values in results.iterrows()
+                for col in results.columns
+            },
+        )
+
+    def get_image_feature_name(self) -> str:
```
nathan-stender, line 149: Again, this accessor should not exist, just use the stored value.

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,369 @@
...
+
+@dataclass(frozen=True)
+class MeasurementInfo:
+    instrument_serial_number: str
+    measurement_time: str
+    protocol_signature: str
+    measurement_signature: str
+
+    @staticmethod
+    def get_instrument_serial_number(elements: dict[str, str]) -> str:
```
nathan-stender, line 175: You should instead follow the `create` pattern: make a single `create` method for Measurement info that gets all these values and returns a `MeasurementInfo`

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
```diff
@@ -0,0 +1,369 @@
...
+    @staticmethod
+    def get_measurement_signature(elements: dict[str, str]) -> str:
+        return assert_not_none(
+            elements.get("Measurement Signature"),
+            msg="Unable to find Measurement Signature in Measurement Information section.",
+        )
+
+
+@dataclass(frozen=True)
+class PlateType:
```
nathan-stender, line 204: This is not used. It is very confusing code. 

From what I can tell, you are using this create method to move past the plate type section of the table.

This should not be done this way. You should move the reader (call the code of this create method) inside the `create_v2_data` method.  I know that means that you can't create the other sections inside the constructor, but that little nice bit of formatting is not worth creating a dummy dataclass that does nothing.

The only reason that it could be argued to keep this is if it is going to be used in the future. In this case, you should add a comment explaining this.

Otherwise, it should be removed.

nathan-stender (changes requested):

src/allotropy/parsers/revvity_kaleido/kaleido_structure.py
alejandro-salgado: The idea of this class was to be used in case some parts of that information were relevant in the future. For the moment I remove that class.

nathan-stender (approved):
LGTM! Thanks for bearing with me on structural changes.

slopez-b (approved):

PR #358, opened 2024-04-01
Add encoding inference to Beckman Vi Cell Blu adapter

alejandro-salgado:
- Reorganize lines reader global functions
- Redefine csv reader methods with kwargs
- Use Csv Reader in Beckman Vi Cell Blu adapter

### DIFF
--- a/src/allotropy/parsers/lines_reader.py
+++ b/src/allotropy/parsers/lines_reader.py
@@ -1,7 +1,7 @@
 from collections.abc import Iterator
 from io import StringIO
 from re import search
-from typing import Any, Literal, Optional, Union
+from typing import Any, Optional
 
 import chardet
 import pandas as pd
@@ -158,19 +158,15 @@ class CsvReader(LinesReader):
 
     def lines_as_df(
         self,
-        *,
         lines: list[str],
-        header: Optional[Union[int, Literal["infer"]]] = None,
-        sep: Optional[str] = ",",
-        as_str: bool = False,
+        **kwargs: Any,
     ) -> Optional[pd.DataFrame]:
         if lines:
             return read_csv(
                 StringIO("\n".join(lines)),
-                header=header,
-                sep=sep,
-                dtype=str if as_str else None,
+                dtype=None,
                 # Prevent pandas from rounding decimal values, at the cost of some speed.
                 float_precision="round_trip",
+                **kwargs,
             )
         return None
--- a/src/allotropy/parsers/lines_reader.py
+++ b/src/allotropy/parsers/lines_reader.py
@@ -14,24 +14,7 @@ from allotropy.named_file_contents import NamedFileContents
 EMPTY_STR_PATTERN = r"^\s*$"
 
 
-def read_to_lines(named_file_contents: NamedFileContents) -> list[str]:
-    stream_contents = named_file_contents.contents.read()
-    encoding = named_file_contents.encoding
-    raw_contents = (
-        _decode(stream_contents, encoding)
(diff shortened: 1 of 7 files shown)


src/allotropy/parsers/lines_reader.py
```diff
@@ -160,19 +158,15 @@ def pop_as_series(self, sep: str = " ") -> Optional["pd.Series[str]"]:
 
     def lines_as_df(
         self,
-        *,
```
nathan-stender, line 163: What is the purpose of removing this?

src/allotropy/parsers/lines_reader.py
```diff
@@ -132,19 +134,15 @@ def pop_csv_block_as_lines(self, empty_pat: str = EMPTY_STR_PATTERN) -> list[str
     def pop_csv_block_as_df(
         self,
         empty_pat: str = EMPTY_STR_PATTERN,
-        *,
```
nathan-stender, line 135: What is the purpose of removing this?

src/allotropy/parsers/lines_reader.py
```diff
@@ -14,34 +14,36 @@
 EMPTY_STR_PATTERN = r"^\s*$"
 
 
-def read_to_lines(named_file_contents: NamedFileContents) -> list[str]:
+def detect_encoding(bytes_content: bytes) -> str:
```
nathan-stender, line 17: I can't tell what the change in logic is here. If there is a change can you clarify what it is? If not, I don't think there is sufficient justification to refactor the code here.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -390,7 +390,7 @@ def create(
 
         # convert rows to df
         data = assert_not_none(
-            reader.lines_as_df(lines=lines, sep="\t"),
+            reader.lines_as_df(lines=lines, sep="\t", header=None),
```
nathan-stender, line 393: Why are all these `header=None` necessary now? that seems like an undesirable change.

nathan-stender (changes requested):

src/allotropy/parsers/lines_reader.py
slopez-b: I second this. Plus, if there's a need to refactor the `lines_reader` which is used by several other adapters, it should have its own PR.

src/allotropy/parsers/lines_reader.py
alejandro-salgado: The idea is to allow the function to pass all the kwargs directly to pandas, instead of the need to add the parameters one by one.

src/allotropy/parsers/lines_reader.py
alejandro-salgado: this is the same idea as in lines_as_df. To stop ignoring parameters and instead allow the function to pass kwargs directly to pandas.

src/allotropy/parsers/lines_reader.py
alejandro-salgado: The idea is to decouple this functions related to read raw files. I can remove this changes from the PR if needed.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
alejandro-salgado: The header=None is necessary if the kwargs approach is implemented. 
However I think that is a good thing that the missing header is explicitly declared.

PR #359, opened 2024-04-01
Initial Biorad Bioplex PR (Structure Only)

dchou-benchling:
- Add bioplex structure and tests
- Add util to values to parse values out of xml


src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+@dataclass(frozen=True)
+class AnalyteSample:
+    analyte_name: str
+    analyte_region: int
+    analyte_error_code: int
+
+    @staticmethod
+    def create(analyte_xml):
```
dchou-benchling, line 41: Note: adding a return definition of "AnalyteSample", results in this error: NameError: name 'AnalyteSample' is not defined

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,261 @@
...
+
+@dataclass
+class SampleDocumentAggregate:
+    # This data class pulled from the <Samples> part of the xml.
+    samples: list[SampleDocument] = field(default_factory=list)
+    # Default to empty dictionary.
+    analyte_region_dict: dict[str, str] = field(default_factory=dict)
+
+    @staticmethod
+    def create(samples_xml):
```
dchou-benchling, line 73: Note: When I add the type for samples_xml (DefusedElementTree.Element), I get the following attribute error: AttributeError: module 'defusedxml.ElementTree' has no attribute 'Element'
This was not happening when using the native ElementTree. This looks related: https://github.com/tiran/defusedxml/issues/48.
I could just go back to the native library and add #noqa?

slopez-b (changes requested):

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,268 @@
...
+    samples: list[SampleDocument] = field(default_factory=list)
+    # Default to empty dictionary.
+    analyte_region_dict: dict[int, str] = field(default_factory=dict)
+
+    @staticmethod
+    def create(samples_xml: ElementTree.Element) -> SampleDocumentAggregate:
+        sample_documents = SampleDocumentAggregate()
+        for sample_types in samples_xml:
+            for child_sample_type in sample_types:
+                # TODO: add catch of non mapping sample roles
```
nathan-stender, line 82: Would you be able to just do this now? You can write a getter func that throws an AllotropeConversionError and call it good :D

src/allotropy/parsers/utils/values.py
```diff
@@ -196,3 +197,20 @@ def num_to_chars(n: int) -> str:
 
 def str_or_none(value: Any) -> Optional[str]:
     return None if value is None else str(value)
+
+
+def get_val_from_xml(
+    xml_object: ElementTree.Element,
+    index_1: int,
```
nathan-stender, line 204: Is there not some way to retrieve these values by name instead of numerical index? 

Looking at the test data, they seem to be labelled. It seems like this would make it much easier to follow the code.

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,268 @@
...
+    sample_volume_setting: int
+    detector_gain_setting: str
+    minimum_assay_bead_count_setting: int
+    well_total_events: int
+    acquisition_time: str
+
+    @staticmethod
+    def create(well_xml: ElementTree.Element) -> DeviceWellSettings:
+        well_name = get_well_name(well_xml.attrib)
+        well_acq_time = get_val_from_xml(well_xml, 2, "AcquisitionTime")
```
nathan-stender, line 156: See comment on get_val_from_xml about using names instead of numerical indexes. If possible, this would be much easier to follow / debug/ modify in the future. Also has the added benefit that you could probably use that for the error message :D

nathan-stender (changes requested):
Overall, looks great! Thank you for the detailed tests!

One request about how we are accessing XML to look into.

src/allotropy/parsers/utils/values.py
nathan-stender: Looks like `xml_object.find` will do this! You can even find values nested inside with a path-like value (e.g. `xml_object.find("name1/name2")`

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
dchou-benchling: Thank you for finding this, being able to not use numerical indexes is huge!

dchou-benchling:
@nathan-stender Incorporated your suggestions, I ended up just using a generic .find() instead of specifying the layers within the xml like I was doing when using indices. I could alternatively use .find("tag1/childtag2"), let me know if you have thoughts

src/allotropy/parsers/utils/values.py
```diff
@@ -201,16 +201,13 @@ def str_or_none(value: Any) -> Optional[str]:
...
-    xml_tag_name: str,
-    index_2: Optional[int] = None,
+    tag_name: str,
 ) -> str:
-    try:
-        if index_2 is None:
-            val_from_xml = xml_object[index_1]
-        else:
-            val_from_xml = xml_object[index_1][index_2]
+    import pdb;pdb.set_trace()
```
slopez-b: remove debugger line

slopez-b (dismissed):

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+    minimum_assay_bead_count_setting: int
+    well_total_events: int
+    acquisition_time: str
+
+    @staticmethod
+    def create(well_xml: ElementTree.Element) -> DeviceWellSettings:
+        well_name = get_well_name(well_xml.attrib)
+        well_acq_time = get_val_from_xml(well_xml, "AcquisitionTime")
+        total_events = get_val_from_xml(well_xml, "TotalEvents")
+        sample_volume = int(get_val_from_xml(well_xml, "SampleVolume"))
```
nathan-stender, line 153:
```suggestion
        sample_volume = try_int(get_val_from_xml(well_xml, "SampleVolume"), "SampleVolume")
```

nathan-stender: (this will get us an helpful error is it's not an int)

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+        analyte_region_dict: dict[str, str],
+        regions_of_interest: list[str],
+    ) -> Optional[AnalyteDocumentData]:
+        # Look up analyte name from sample
+        assay_bead_identifier = bead_region_xml.attrib[REGION_NUMBER]
+        # Look up bead region -> analyte name
+        if assay_bead_identifier in regions_of_interest:
+            analyte_name = analyte_region_dict[assay_bead_identifier]
+            # Region Count is first element of bead region.
+            assay_bead_count = int(get_val_from_xml(bead_region_xml, "RegionCount"))
```
nathan-stender, line 188: try_int

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+            # Median
+            fluorescence = float(get_val_from_xml(bead_region_xml, "Median"))
```
nathan-stender, line 190: try_float

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+        # This is the RunProtocolDocumentLocation
+        analytical_method = get_val_from_xml(xml_well, "RunProtocolDocumentLocation")
+        regions = xml_well[6][4]
+        if regions.tag != REGIONS_OF_INTEREST:
+            msg = f"Could not find {REGIONS_OF_INTEREST} in instrument xml"
+            raise AllotropeConversionError(msg)
+
+        regions_of_interest = []
+        for region in regions:
+            int_region = int(region.attrib[REGION_NUMBER])
```
nathan-stender, line 224: try_int

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+class AnalyteSample:
+    analyte_name: str
+    analyte_region: int
+    analyte_error_code: int
+
+    @staticmethod
+    def create(analyte_xml: ElementTree.Element) -> AnalyteSample:
+        return AnalyteSample(
+            analyte_name=str(analyte_xml[0].text),
+            analyte_region=int(analyte_xml.attrib[REGION_NUMBER]),
```
nathan-stender, line 49: try_int

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+            analyte_error_code=int(analyte_xml[1].attrib[CODE]),
```
nathan-stender, line 50: try_int

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+    @staticmethod
+    def create(well_xml: ElementTree.Element) -> DeviceWellSettings:
+        well_name = get_well_name(well_xml.attrib)
+        well_acq_time = get_val_from_xml(well_xml, "AcquisitionTime")
+        total_events = get_val_from_xml(well_xml, "TotalEvents")
+        sample_volume = int(get_val_from_xml(well_xml, "SampleVolume"))
+        # RP1 Gain. 16th element of Run Conditions
+        detector_gain_setting = get_val_from_xml(well_xml, "RP1Gain")
+        # Assuming second element of Run Settings
+        # Can't use util here because this pulls from attrib
```
nathan-stender, line 157: You could factor out the part that gets the element (e.g. `get_element_from_xml`) to get the element, then have `get_val_from_xml` call this to get the element, and then access the text.

Then, you can either make a `get_attribute_from_xml` that calls `get_element_from_xml` and then gets `.attrib[ATTRIBUTE_NAME]`. Or, if it's only used here, I'm find with you just calling the attrib part here.

The part I really want to avoid is hard-coded indices.

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
nathan-stender: I tried to note other places in code that I noticed bare int/float conversions to apply this same pattern.

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+        min_assay_bead_count_setting = well_xml[6][1].attrib[BEAD_COUNT]
+        return DeviceWellSettings(
+            well_name=well_name,
+            sample_volume_setting=sample_volume,
+            detector_gain_setting=detector_gain_setting,
+            minimum_assay_bead_count_setting=int(min_assay_bead_count_setting),
```
nathan-stender, line 163: try_int

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+            acquisition_time=well_acq_time,
+            well_total_events=int(total_events),
```
nathan-stender, line 165: try_int

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
```diff
@@ -0,0 +1,272 @@
...
+                    sample_dilution = get_val_from_xml(child_sample_type, "Dilution")
+                    for child in child_sample_type:
+                        if child.tag == MEMBER_WELLS:
+                            for member_well in child:
+                                sample_document = (
+                                    SampleDocumentAggregate._generate_sample_document(
+                                        sample_documents=sample_documents,
+                                        well_xml=member_well,
+                                        sample_id=sample_identifier,
+                                        sample_dilution=float(sample_dilution),
```
nathan-stender, line 98: try_float

nathan-stender (changes requested):

src/allotropy/parsers/utils/values.py
```diff
@@ -196,3 +197,16 @@ def num_to_chars(n: int) -> str:
 
 def str_or_none(value: Any) -> Optional[str]:
     return None if value is None else str(value)
+
+
+def get_val_from_xml(
+    xml_object: ElementTree.Element,
+    tag_name: str,
+) -> str:
+    val_from_xml = xml_object.find(".//" + tag_name)
```
nathan-stender, line 206: I think it would be best to require a full path (as you note in your comment to me)

This reduces chance of an error occurring because there is a different, unexpected tag with the same name elsewhere in the document.

src/allotropy/parsers/biorad_bioplex/biorad_bioplex_structure.py
nathan-stender: Didn't fix this one

nathan-stender (changes requested):
Just a couple more places!

nathan-stender (approved):
🎉

PR #369, opened 2024-04-09
add AppBio Quantstudio Design and Analysis adapter

alejandro-salgado:
- Add AppBio Quantstudio Design and Analysis adapter


src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_contents.py
```diff
@@ -0,0 +1,68 @@
...
+from allotropy.parsers.utils.values import (
+    assert_not_empty_df,
+    assert_not_none,
+)
+from allotropy.types import IOType
+
+
+class DesignQuantstudioContents:
+    def __init__(self, contents: IOType) -> None:
+        self.raw_contents = pd.read_excel(contents, header=None, sheet_name=None)
```
nathan-stender, line 17: contents and raw_contents are not accessed after they are used to create header and data, they should not be stored as member variables. Instead they should be normal variables, and should be passed into _get_header and _get_data

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_structure.py
```diff
@@ -0,0 +1,569 @@
...
+            ),
+            model_number=try_str_from_series_or_none(header, "Instrument Type") or "NA",
+            device_serial_number=(
+                try_str_from_series_or_none(header, "Instrument Serial Number") or "NA"
+            ),
+            measurement_method_identifier=try_str_from_series(
+                header, "Quantification Cycle Method"
+            ),
+            pcr_detection_chemistry=(
+                try_str_from_series_or_none(header, "Chemistry") or "N/A"
```
nathan-stender, line 109: We are using "NA" above, but "N/A" here, which is correct?

src/allotropy/parsers/utils/values.py
```diff
@@ -183,7 +191,7 @@ def try_bool_from_series_or_none(
 ) -> Optional[bool]:
     try:
         value = data.get(key)
-        return None if value is None else bool(value)
+        return None if value is None else str(value).lower() == "true"
```
nathan-stender, line 194: Ah, I should have caught this originally.  Can you add a function like this:

```
def str_to_bool(v):
  return v.lower() in ("yes", "true", "t", "1")
``` 

And call that instead?

```
return None if value is None else str_to_bool(value)
```

tests/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_parser_test.py
```diff
@@ -0,0 +1,36 @@
...
+    "appbio_quantstudio_designandanalysis_QS7_Standard_Curve_example06",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Genotyping_example07",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Multiplex_example08",
+    "appbio_quantstudio_designandanalysis_QS7Pro_PCR_with_Melt_example09",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Presence_and_Absence_example10",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Relative_Quantification_example11",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Relative_Quantification_Biogroup_example12",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Relative_Standard_Curve_example13",
+    "appbio_quantstudio_designandanalysis_QS7Pro_Standard_Curve_example14",
+    ## "appbio_quantstudio_designandanalysis_QS7Pro_Standard_Curve_TAC_example15",
```
nathan-stender, line 21: Add comment on why these tests are excluded, or remove?

nathan-stender (changes requested):
Great PR! A couple changes

tests/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_parser_test.py
alejandro-salgado: Those are files not supported in this version, but they should be supported in the next one

nathan-stender (approved):

nathan-stender (approved):

PR #372, opened 2024-04-10
chore: Add missing CHANGE_NOTES for multi-analyte

nathan-stender:


LimulusPolyphemus (changes requested):
@nathan-stender actually the initial version of this schema was published by Allotrope as part of the 2024/03 release.

https://gitlab.com/allotrope/adm/-/blob/develop/purl/json-schemas/adm/multi-analyte-profiling/REC/2024/03/multi-analyte-profiling.schema.json

Therefore, this would represent the base schema.

Also, the `Calibration Document` was not implemented per our proposal, so that is an extension present in our schema that is not in the base schema.

nathan-stender:
Oh yeah, I noticed that and didn't mean to mark this as ready for review, I was gonna ask you about the diffs to note 🤣

PR #376, opened 2024-04-12
chore: Add proposal/revisions to CHANGE NOTES for qPCR and cell-counting schemas

james-leinas:


nathan-stender (changes requested):
Sync'd offline - the purpose of CHANGE_NOTES is to track changes between the BENCHLING schema and the base REC ASM schema, not to propose new changes to the BENCHLING schema

PR #379, opened 2024-04-15
fix: Enable encoding detection in vi-cell-blu

nathan-stender:


src/allotropy/parsers/lines_reader.py
```diff
@@ -36,11 +36,14 @@ def _determine_encoding(bytes_content: bytes, encoding: Optional[str]) -> str:
     if not detected:
         error = "Unable to detect text encoding for file. The file may be empty."
         raise AllotropeConversionError(error)
+    # Windows-1252 is a subset of UTF-8, and may lead to missing some data.
```
mdlu, line 39: I'd be okay with this for now, but I'm wondering how safe this is / why it's not being picked up as UTF-8? the documentation suggests that it just defaults to Windows-1252 when it can't figure it out (https://chardet.readthedocs.io/en/latest/how-it-works.html) so I guess this effectively makes the "default" UTF-8 but just some food for thought / wondering if you have more insight here

mdlu: also is there a particular file that's been giving you issues where you could add a test case (where it detects as `Windows-1252` when it should be `utf-8`)?

mdlu (approved):

src/allotropy/parsers/lines_reader.py
nathan-stender: Yeah idk why it defaults to `Windows-1252` when `UTF-8` is the seeming standard everywhere.

We do have a test case covering this though, the VI-Cell blu example 2, which has an emoji in the file.  I expect is was explicitly added to catch encoding issues, and here, when we use CHARDET, it fails due to returning `Windows-1252`, and changing that to `UTF-8` fixes the issue.

mdlu (approved):

PR #380, opened 2024-04-16
Chore: Correct Bio-Plex Manager Test Files

LimulusPolyphemus:
Corrected Bio-Plex Manager test files to remove potential identifiers


nathan-stender (commented):
@LimulusPolyphemus looks like the test file changes require corresponding changes in the parser? We won't be able to merge the change with failing tests.

I think you'll either need to have someone make the fix and merge with the test file fixes, or add these as new file names (e.g. file_name_fixed) and then when someone makes the corresponding fixes they can replace the existing files

LimulusPolyphemus:
@nathan-stender can you tell what it is causing the failure? The only changes should have been to string values which are expected to vary.

The only other potential change was that when I regenerated the .json file, the "mu" character was being rendered as the ASCII key rather than the native character. However, I changed the ASCII back to native the test was still failing.

nathan-stender:
@LimulusPolyphemus we validate the output contents exactly, so changing a string value will break the test. If there are no functional changes, you prob just need to regenerate the test results. 

I can take a crack at that later today

LimulusPolyphemus:
@nathan-stender I did regenerate the JSON file using the adapter after correcting the XML test file - that's why I'm confused as to why the test is failing.

nathan-stender:
@LimulusPolyphemus I'm not sure what happened for you, but I regenerated the test files in a branch and got different results, pushing those up.

nathan-stender:
@LimulusPolyphemus actually I just realized, the problem was that for tests we have the file use a faked ID generator, so that we get the same result every time. If you generated without this, the IDs would be different, which would cause the diff.

nathan-stender (approved):

LimulusPolyphemus:
@nathan-stender ahh, that makes sense. I had forgotten about that.

mdlu (approved):

PR #381, opened 2024-04-18
Biorad Bioplex Manager revisions

dchou-benchling:
- Rename the Biorad Bioplex to "Biorad Bioplex Manager"
- Add a utility to remove non-required None values from a dataclass
- Use ASM_CONVERTER_NAME specific to adapter, comment out line where
- Remove description in ASM output when "None"


src/allotropy/parsers/utils/values.py
```diff
@@ -256,3 +260,31 @@ def get_attrib_from_xml(
     except KeyError as e:
         msg = f"Unable to find '{attrib_name}' in {xml_element.attrib}"
         raise AllotropeConversionError(msg) from e
+
+
+def remove_none_fields_from_data_class(
```
nathan-stender, line 265: Can we get some unit tests for this?

nathan-stender (commented):
Code LGTM - requesting tests on `remove_none_fields_from_dataclass`. Be sure to cover case with required field that defaults to None :D

nathan-stender (changes requested):
Whoops meant to request changes on last comment.

src/allotropy/parsers/utils/values.py
dchou-benchling: @nathan-stender I added tests but had to ignore the incompatible type mypy errors when using `Optional[str]` since `make_dataclass` expected `type` not `object`

nathan-stender: No worries! 

I _think_ you could get around this by just defining the dataclasses outside of the test function.

But also, I don't really care about type ignores in a unit test :D

nathan-stender (approved):

nathan-stender (approved):

PR #383, opened 2024-04-19
add calculated data documents to AppBio Quantstudio Design and Analysis

alejandro-salgado:
- add calculated data documents to AppBio Quantstudio Design and Analysis


src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -0,0 +1,677 @@
...
+    if r_ct_sd_ref is None:
+        return None
+
+    return CalculatedDocument(
+        uuid=random_uuid_str(),
+        name="delta equivalent ct sd",
+        value=delta_ct_sd,
+        data_sources=[
+            DataSource(feature="ct sd", reference=ct_sd_ref),
+            DataSource(feature="ct sd", reference=r_ct_sd_ref),
```
nathan-stender, line 262: Just confirming - we don't signify which value is reference value? I'm confused how you can tell from the data document which value is which.

nathan-stender (commented):
Code looks good, one question on data source identifiers

@LimulusPolyphemus have you reviewed the calculated data results to confirm they look right? I don't have any context to say the values are correct, except to check that values are properly linked, etc.

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
alejandro-salgado: Yes, the feature is the same. The way that the reference is determined is with the data_processing_document when the calculated data documents are constructed in the schema.

LimulusPolyphemus: @nathan-stender @alejandro-salgado - It is hard to check all of these manually based on the JSON - but the `calculated data documents` and their associated `data source documents` look okay to me.

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -0,0 +1,677 @@
...
+) -> Optional[CalculatedDocument]:
+    well_items = view_data.get_leaf_item(sample, target)
+    if (quantity_mean := well_items[0].result.quantity_mean) is None:
+        return None
+
+    data_sources = []
+    for well_item in well_items:
+        quantity_ref = build_quantity(well_item)
+        if quantity_ref is None:
+            return None
```
nlangowitz, line 44: Are we sure this is the correct behavior if even 1 well_item has null quantity? Should there be any more detailed logging if we get a document where `quantity_mean` above is not None, but a `quantity_ref`is?

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -0,0 +1,677 @@
...
+
+    return CalculatedDocument(
+        uuid=random_uuid_str(),
+        name="quantity mean",
+        value=quantity_mean,
+        data_sources=data_sources,
+    )
+
+
+def build_quantity_sd(
```
nlangowitz, line 61: Is this exactly the same as the above function, just with "quantity sd" as the name and referencing `well_items[0].result.quantity_sd` instead of `well_items[0].result.quantity_mean`? Doesn't need to be this PR, but I think it'd benefit from a refactor to a shared `build_quantity_calculated_document(view_data, sample, target, name, get_calculation: Callable[[WellItem], float])` (or `calculation_property: str` but that gives up some type safety)

nlangowitz (approved):
✅ given Nate had only the one question before and the answer to that makes sense, left a few nonblocking suggestions/questions

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
LimulusPolyphemus: @nlangowitz thanks - we'll tackle this and the recommendation below in another PR.

PR #384, opened 2024-04-22
Use new plate reader schema in gen5 adapter

slopez-b:
Use the `BENCHLING/2023/09/plate-reader` schema for the gen5 adapter.

This PR includes the parsing of: 

- Ultraviolet Absorbance gen5 files
- Fluorescence gen5 files
- Luminescence gen5 files

Out of scope of this PR:
- Imaging datapoint files will be handled in it's own parser due to the formatting differences of those files and the schema.


src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+        return ReadData(
+            read_mode=read_mode,
+            step_label=device_control_data.get("Step Label"),
+            measurement_labels=measurement_labels,
+            detector_carriage_speed=device_control_data.get(READ_SPEED_KEY),
+            # Absorbance attributes
+            pathlength_correction=device_control_data.get(PATHLENGTH_CORRECTION_KEY),
+            number_of_averages=try_float_or_none(number_of_averages),
+            # Luminescence attributes
+            detector_distance=(try_float_or_none(read_height.split(" ")[0])),
```
nathan-stender, line 211:
```suggestion
            detector_distance=try_float_or_none(read_height.split(" ")[0]),
```

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+
+            elif strp_line.startswith(WAVELENGTHS_KEY):
+                wavelengths = strp_line.split(":  ")
+                read_data_dict[WAVELENGTHS_KEY].extend(wavelengths[1].split(", "))
+                continue
+
+            line_data: list[str] = strp_line.split(",  ")
+            for read_datum in line_data:
+                splitted_datum = read_datum.split(": ")
+                if len(splitted_datum) != datum_len:
```
nathan-stender, line 324:
```suggestion
                if len(splitted_datum) != 2:
```

There isn't a good reason for pulling this out into a "constant", it seems to make the code less clear to me. If lint was complaining about this, ignore the rule on this line and I'll investigate.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+                else:
+                    read_data_dict[splitted_datum[0]] = splitted_datum[1]
+
+        return read_data_dict
+
+    @classmethod
+    def _get_step_label(cls, read_line: str, read_mode: str) -> Optional[str]:
+        read_data_len = 2
+        split_line = read_line.split("\t")
+        if len(split_line) != read_data_len:
```
nathan-stender, line 337:
```suggestion
        if len(split_line) != 2:
```
Same comment as above.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+        return LayoutData(sample_identifiers={})
+
+    @staticmethod
+    def create(layout_str: str) -> LayoutData:
+        identifiers = {}
+
+        layout_lines: list[str] = layout_str.splitlines()
+        # first line is "Layout", second line is column numbers
+        current_row = "A"
+        for i in range(2, len(layout_lines)):
```
nathan-stender, line 394: This data is in a tabular format, it may be better to read this in as a csv block using pandas?

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+        read_data: ReadData,
+    ) -> None:
+        result_lines = results.splitlines()
+        calculated_data: defaultdict[str, list] = defaultdict(list)
+        if result_lines[0].strip() != "Results":
+            msg = f"Expected the first line of the results section '{result_lines[0]}' to be 'Results'."
+            raise AllotropeConversionError(msg)
+
+        # result_lines[1] contains column numbers
+        current_row = "A"
```
nathan-stender, line 479: Also here - can we read lines as a csv?

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
slopez-b: yes, this was a complain by the linter... I agree with you it is not necessary to use a variable for this.

tests/parsers/agilent_gen5/testdata/fluorescence/endpoint_singleplate.txt
```diff
@@ -19,7 +19,7 @@ Reading Type	Manual
 Procedure Details
 
 Plate Type	96 WELL PLATE (Use plate lid)
-Eject plate on completion
```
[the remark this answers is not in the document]
slopez-b: No, it works either way, I think my autoformatter removed the trailing whitespace somehow. This line is even ignored in the parser. I can change it back if you want.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+class HeaderData:
+    software_version: str
+    experiment_file_path: str
+    protocol_file_path: str
+    datetime: str
+    well_plate_identifier: str
+    model_number: str
+    equipment_serial_number: str
+
+    @classmethod
```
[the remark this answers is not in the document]
slopez-b: I use `classmethods` mainly for two reasons:

- It helps maintainability. Let's suppose the name of the class changes, you wouldn't have to change the method calls everywhere if you're use classmethods. Also, if we are calling the methods using the classname everywhere it is more prone to error and typos.
- These are actually no static methods. Those methods are not supposed to be called outside this create method, that's why they are protected in the first place. It does not make much sense to have protected methods that are static at the same time.

Let me know what you think.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -1,26 +1,574 @@
...
+        elif ReadType.SPECTRAL.value in procedure_details:
+            return ReadType.SPECTRAL
+        # check for this last, because other modes still contain the word "Endpoint"
+        elif ReadType.ENDPOINT.value in procedure_details:
+            return ReadType.ENDPOINT
+
+        msg = f"Read type not found; expected to find one of {sorted(ReadType._member_names_)}."
+        raise AllotropeConversionError(msg)
+
+    @classmethod
```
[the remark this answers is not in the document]
slopez-b: see above explanation.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
slopez-b: TBH, I inherited this from the old gen5 adapter and didn't think on changing it because it was working fine. Will look into it.

slopez-b: Given the way the data is formatted (it can be multidimentional) and the fact that we would still have to iterate through all rows and columns to get the well positions. I think we wouldn't gain much by reading it as a csv with pandas.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
slopez-b: see above.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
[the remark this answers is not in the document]
nathan-stender: Sounds good. I don't find the first reason (maintainability) compelling, as I don't think we should be renaming classes enough to justify confusing static/classmethod divide. However, I do agree with the second reason (protected methods).

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
nathan-stender: I think it would help with readability, as you should be able to get some column headers, but given that you inherited this code and this is already a pretty big PR, I'm good with leaving it 👍

nathan-stender (dismissed):
Code LGTM - please get approval from both @alejandro-salgado and @LimulusPolyphemus for calculated data documents data as well.

alejandro-salgado (dismissed):
Code for calculated data documents LGTM.

alejandro-salgado (approved):

PR #389, opened 2024-04-25
Add Display Names for Vendors

stephenworlow:
Add Display Names for Vendors


LimulusPolyphemus (changes requested):
@stephenworlow could we change "Biorad Bioplex" to "BioRad BioPlex Manager"? There is another application called "BioRad BioPlex Data Pro" and I want to be sure we're distinguishing between the two.

LimulusPolyphemus (changes requested):
@stephenworlow one more, could we change:
"AppBio QuantStudio" --> "AppBio QuantStudio RT-PCR"
"Luminex XPonent" --> "Luminex xPONENT"
"Qiacuity DPCR" --> "Qiacuity dPCR"

nathan-stender (approved):

LimulusPolyphemus (approved):

PR #396, opened 2024-05-03
add cache to calculated data constructors in AppBio design Quantstudio

alejandro-salgado:
add cache to calculated data constructors in design quantstudio

This functions are design to generate a CalculatedData object each time they are called if the data they represent is available.

The CalculatedData object contains an uuid. 

So, assuming the corresponding data is available, if the function is invoked two times with the same parameters, two CalculatedData objects will be created and their only difference would be the uuid.

With the cache decorator there would not be duplicated CalculatedData objects.

This PR modifies the JSON outputs because it removes the duplicated calculated documents that were present because of the previous absence of cache in these functions.


nathan-stender (changes requested):
This is not something that I would expect to change test results.

If this expected, ok, can you please explain what was wrong before and why this is right in the PR description?

nathan-stender (approved):

PR #397, opened 2024-05-04
Improve AppBio Quantstudio tests

alejandro-salgado:
- add mark to quantstudio tests
- remove duplicated test file


tests/parsers/appbio_quantstudio/appbio_quantstudio_structure_test.py
```diff
@@ -59,14 +59,15 @@ def _read_to_lines(io_: IOType, encoding: Optional[str] = None) -> list[str]:
     return read_to_lines(named_file_contents)
 
 
-@pytest.mark.short
+@pytest.mark.quantstudio
```
nathan-stender, line 62: You can add multiple labels to a single test. I think we should keep the "short" label, for when a user wants to run all tests that arent too long, but I'm fine with adding the "quantstudio" tag as well.

nathan-stender (changes requested):

slopez-b (approved):

PR #399, opened 2024-05-04
remove inner calculated data documents from AppBio Quantstudio

alejandro-salgado:
- remove inner calculated data documents from AppBio Quantstudio

This change requires the refactor of the PCR CSV converter. This task is already being developed.


nathan-stender (approved):
Approving this, but perhaps we should wait to merge until the CSV parser code is ready?

PR #405, opened 2024-05-10
add first version of tool to create graph visualization of calculated data documents

alejandro-salgado:
- add tool to create graph visualization of calculated data documents


src/allotropy/parsers/utils/calculated_data_documents/visualization.py
```diff
@@ -0,0 +1,66 @@
+from typing import Any
+
+import networkx as nx  # type: ignore[import-not-found]
```
slopez-b, line 3: @alejandro-salgado why are we ignoring this import warning?

PR #408, opened 2024-05-14
Expand PharmSpec tests

chrisfriedline-benchling:
This PR adds more testing for the PharmSpec adapter. The files include varying numbers of runs, with and without calculated docs.


tests/parsers/beckman_pharmspec/beckman_pharmspec_parser_test.py
```diff
@@ -22,129 +22,223 @@
 
 
 @pytest.fixture()
-def test_file() -> Path:
+def test_file_1() -> Path:
```
nathan-stender, line 25: Can we replace these with a list of test file names, and a couple helpers to get paths from them? Similar to other test files.

See `vi_cell_blu_parser_test.py` as one example.

tests/parsers/beckman_pharmspec/beckman_pharmspec_parser_test.py
```diff
@@ -22,129 +22,223 @@
...
+        "test_file_3": test_file_3,
+        "test_file_4": test_file_4,
+        "test_file_5": test_file_5,
+    }
+
+
 @pytest.mark.short
-def test_get_model(test_file: Path) -> None:
-    parser = PharmSpecParser(TimestampParser())
+def test_get_model(test_files: dict[str, Path]) -> None:
```
nathan-stender, line 92: As part of the request to change how we are passing test files, you can use parametrize to create test cases. `unchained_labs_lunatic_structure_test` has an example of multi parameters.

Something like:

```
@pytest.mark.parametrize(
    "file_name,num_measurement_docs,has_calculated_doc",
    [
        (OUTPUT_FILES[0], 2, True),
        ...
     ]
)

**Review by nathan-stender — CHANGES_REQUESTED** [review]

File: tests/parsers/beckman_pharmspec/beckman_pharmspec_parser_test.py (hunk 2 above)
Comment (chrisfriedline-benchling) [reply tests/parsers/beckman_pharmspec/beckman_pharmspec_parser_test.py]: Fixed

File: tests/parsers/beckman_pharmspec/beckman_pharmspec_parser_test.py [hunk 3]
```diff
@@ -133,17 +155,30 @@ def test_get_model(test_file: Path) -> None:
...
+    else:
+        assert (
+            not model.light_obscuration_aggregate_document.calculated_data_aggregate_document
+        )
+
 
 @pytest.mark.short
-def test_asm(test_file: Path) -> None:
+@pytest.mark.parametrize("file_name", OUTPUT_FILES)
+def test_asm(file_name: str) -> None:
```
nathan-stender, line 170: This isn't new in your PR, but we should remove this test, as it's redundant.

The test below calls `allotrope_from_file` and actually checks contents, while this just checks that we could process the file at all.

nathan-stender (changes requested):

tests/parsers/beckman_pharmspec/beckman_pharmspec_parser_test.py
chrisfriedline-benchling: Done

nathan-stender (approved):

nathan-stender:
Went ahead and made a fresh PR to add these here: https://github.com/Benchling-Open-Source/allotropy/pull/678

PR #409, opened 2024-05-15
add first version of tool to create graph visualization of calculated data documents

alejandro-salgado:
- add first version of tool to create graph visualization of calculated data documents


pyproject.toml
```diff
@@ -118,6 +118,8 @@ extra-dependencies = [
   "pandas-stubs == 2.1.4.231227",
   # For pytest types
   "pytest >= 7.4.0",
+  # For check graph generation
+  "networkx-stubs",
```
slopez-b, line 122: @alejandro-salgado shouln't this requirement be in the `tests` or `default` env settings instead of the `lint` settins?

alejandro-salgado: this script is not for execution at the moment, this is is part of the lint environment only because is the only one that  interacts with the code for the moment.

In the following PR a hatch script to execute the graph visualization will be added and the dependency to the networkx library will be added to it.

slopez-b (approved):

PR #410, opened 2024-05-15
structure for methodical mind

dchou-benchling:


src/allotropy/parsers/methodical_mind/methodical_mind_structure.py
```diff
@@ -0,0 +1,169 @@
...
+class CombinedData:
+    file_name: str
+    version: str
+    model: str
+    serial_number: str
+    plate_doc_info: list[PlateData]
+
+    @staticmethod
+    def create(reader: CsvReader) -> CombinedData:
+        file_name = assert_not_none(reader.pop_if_match(FILENAME), FILENAME).split(
```
stephenworlow, line 31: Nit: I think you could split this out into a method like

`
def get_parameter(reader: CsvReader, name: str) -> str:
       return reader.drop_until_inclusive(name).split("\t")[1]
`

stephenworlow (dismissed):

src/allotropy/parsers/methodical_mind/methodical_mind_structure.py
dchou-benchling: That makes sense, I can do that. Thanks!

stephenworlow (approved):

PR #411, opened 2024-05-15
Add script to create graph visualization of calculated data documents from asm json files

alejandro-salgado:
This PR is the next step of https://github.com/Benchling-Open-Source/allotropy/pull/409

The idea is to build an script that can be executed with hatch to use the graph visualization code of the previous PR

in this case the execution would be **hatch run scripts:visualization \<asm file\> \<calc document name\>**


scripts/visualization.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    parser.add_argument(
+        "input_file",
+        type=str,
+        help="asm json file to visualize",
+    )
+    parser.add_argument(
+        "calculated_document_type",
+        type=str,
+        help="type of calculated data document to visualize",
```
slopez-b, line 87: Could you add some examples of the type of calculated documents that can be passed to this argument? Or is this actually the name of the calculated data document that will be visualized?

alejandro-salgado: The names are the **Calculated data name**, for example **rq min, ct sd, y-intercept, slope**

slopez-b: Nitpick: I think `calculated_document_type` is a bit misleading as a name for the argument. I'd suggest something like `calculated_data_name` or `calculated_document_name`. Just a suggestion.

slopez-b (dismissed):

stephenworlow (dismissed):

slopez-b (approved):

PR #414, opened 2024-05-16
Methodical Mind: Account for params that are blank in the instrument export

dchou-benchling:
For now, this only applies to the "User" field that may or may not be blank.


src/allotropy/parsers/methodical_mind/methodical_mind_structure.py
```diff
@@ -95,23 +97,41 @@ def create(reader: CsvReader) -> CombinedData:
         )
 
     @staticmethod
-    def _get_parameter(reader: CsvReader, name: str) -> str:
-        return assert_not_none(reader.drop_until_inclusive(name), name).split("\t")[1]
+    def _get_parameter_required(
```
stephenworlow, line 100: Nit: you could write this as
```
def get_parameter(read: CsvReader, name: str, is_required: Optional[bool]=True):
    line = assert_not_none(reader.drop_until_inclusive(name), name).split("\t"))
    content = line[1] if len(line) > 2 else None
     if is_required and not content:
     return content
```
And then update above `user = CombinedData._get_parameter(reader, USER, is_required=False)`

dchou-benchling: Good idea, that should work. Thank you!

stephenworlow (approved):

nathan-stender (approved):

dchou-benchling:
Closing this, opened this https://github.com/Benchling-Open-Source/allotropy/pull/420 due to issues with commit gpg signing

PR #421, opened 2024-05-22
Rework PharmSpec Calculated Data and Distribution Identifiers

chrisfriedline-benchling:
This PR assigns IDs for both the calculated data and the distributions. Having IDs in both places allows the calculated data to be associated to the distribution data from which it derives.


src/allotropy/parsers/beckman_pharmspec/beckman_pharmspec_parser.py
```diff
@@ -259,4 +263,79 @@ def _setup_model(self, df: pd.DataFrame, file_name: str) -> Model:
...
+                        )
+                        distribution_id = (
+                            self._get_distribution_id_for_run_and_particle_size(
+                                run_name, float(particle_size), measurement_doc_items
+                            )
+                        )
+                        if distribution_id is not None:
+                            dsd.data_source_identifier = distribution_id
+
+    def _setup_model(self, df: pd.DataFrame, file_name: str) -> Model:
```
nathan-stender, line 317: I'm not going to block this PR on this, but a request.

This parser is getting complicated enough that it probably makes sense to break out this prep work into a set of structure classes.

Generally - we want the parser class to be responsible for calling out to other classes that read the data and using them to construct the model.

We skip the structure pattern when a parser is really simple, but this is going to the point where I think it should be factored out.

nathan-stender (approved):

src/allotropy/parsers/beckman_pharmspec/beckman_pharmspec_parser.py
chrisfriedline-benchling: Makes sense. I'll follow-up.

nathan-stender (dismissed):

nathan-stender (approved):

PR #428, opened 2024-05-23
Add CTL Immunospot adapter

alejandro-salgado:
- Add CTL Immunospot adapter


src/allotropy/parsers/ctl_immunospot/ctl_immunospot_structure.py
```diff
@@ -0,0 +1,243 @@
...
+    def create(reader: LinesReader) -> DeviceInfo:
+        data = {}
+        for i in range(3):  # for first three lines
+            line = assert_not_none(
+                reader.pop(),
+                msg=f"Unable to read line number {i+1}",
+            )
+            for element in line.split(";"):
+                parts = element.split(":", maxsplit=1)
+                if len(parts) > 1:
```
nathan-stender, line 38: Sanity checking: if this is not true - would we be missing expected data, or is that a valid use case. If we expect it to always be true, we should raise an error.

src/allotropy/parsers/ctl_immunospot/ctl_immunospot_structure.py
```diff
@@ -0,0 +1,243 @@
...
+                    data[parts[0].strip()] = parts[1].strip()
+
+        path = assert_not_none(
+            reader.pop(),
+            msg="Unable to read file path",
+        )
+
+        basename = ntpath.basename(path)
+        file_name, *_ = basename.split(".", maxsplit=1)
```
nathan-stender, line 47: Can we use pathlib instead of ntpath here?

Motivating example, you can re-write these two lines as:

file_name = Path(path).stem

src/allotropy/parsers/ctl_immunospot/ctl_immunospot_structure.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def iter_plates_well(self, pos: str) -> Iterator[Well]:
+        for plate in self.plates:
+            yield plate.get_well(pos)
+
+
+@dataclass
+class Data:
+    device_info: DeviceInfo
+    assay_info: AssayInfo
```
nathan-stender, line 227: This doesn't appear to be used anywhere, and you're not really doing any parsing of the lines?

If this will be used in a follow-up PR, then you can just add a comment here saying use will be implemented soon. 

If it is not used, you should remove it, otherwise, you should just skip over it in the create method by calling the `drop_until_empty` method with a comment about skipping over these lines.

nathan-stender (changes requested):
Code looks good! A few requests

src/allotropy/parsers/ctl_immunospot/ctl_immunospot_structure.py
alejandro-salgado: The data that we want to capture is always evaluated as true with this condition. 
There might be other information present that is simply not needed and we want to ignore it.

src/allotropy/parsers/ctl_immunospot/ctl_immunospot_structure.py
alejandro-salgado: I tried to use path at first, but the path might be on posix or windows format.
With ntpath both cases are handled correctly.

nathan-stender: In this case, you can use `PureWindowsPath` from pathlib. I will support both Windows and Posix for this use case (parsing out filename)

nathan-stender (changes requested):
Replied to comment suggesting PureWindowsPath, after that LGTM

PR #430, opened 2024-05-23
fix: use optical-imaging as default device type on gen5-image adapter

slopez-b:
Use optical-imaging as default device type on gen5-image adapter


nathan-stender (changes requested):
Change looks good, please add to CHANGELOG and request re-review

slopez-b:
> Change looks good, please add to CHANGELOG and request re-review

@nathan-stender I did not add anything to CHANGELOG since this is part of the Agilent Gen5 Image adapter which is still unreleased (already in the changelog).

nathan-stender (approved):
Fair enough 👍

PR #434, opened 2024-05-28
feat: Add display_name and release_state as properties on VendorParser

nathan-stender:
This PR changes the strategy for display_name to require `VendorParser` to implements it.

This removes the a touchpoint that could be forgotten for adding a display name for a parser.

It also adds a new property `is_ready_to_use` that the `VendorParser` also implements, which indicates if the parser is ready to be used by customers.

Finally, it enforces that all vendors marked as ready to use are listed in the README with their display names.


stephenworlow (approved):

LimulusPolyphemus:
@nathan-stender just a thought, but rather than 'is_ready_to_use' should we consider adopting the maturity levels that Allotrope uses for the schemas themselves:

- REC - Recommended
- CR - Candidate Release
- WD - Working Draft

It would seem the 'is_recommended' could serve the same function as 'is_ready_to_use'.

LimulusPolyphemus (approved):

PR #443, opened 2024-06-04
chore: Update schema paths

nathan-stender:
This PR changes the way we structure generated schema models to use a directory structure.

Note that there are no functional changes for parsers in this PR. Other than changes to the schema model generation library to work with the new file structure, all changes are refactoring.

It also changes a couple other details about how we store schemas, to bring them more in line with the allotrope repo. Specifically:

1. Add `.schema` suffix to schema files, so `plate_reader.json` -> `plate_reader.schema.json`
2. Nest all schemas under `adm/` root in `schemas` directory.

This is another change in preparation for modular schemas.

We could keep the model directory flat, but that would lead to a huge number of files in a single directory once we are creating model files for every modular schema. The reason we originally used the flat structure was because some folders in the schema layout are numbers, which don't work with python modules. The work-around is to use underscores, e.g. `/2024/09` -> `/_2024/_09` -> `import ..._2024._09...`. While this isn't the most pretty, it's worth it.

We add the `.schema` suffix simply to match the allotrope repo.

We nest under `adm/` because some core schemas that we will download when implementing modular schemas are not under `adm/` but `qudt/`, so we will want this level.


src/allotropy/allotrope/schema_parser/generate_schemas.py
```diff
@@ -84,45 +85,56 @@ def _generate_schema(model_path: Path, schema_path: Path) -> None:
...
-
-    :root_dir: The root directory of the project.
     :dry_run: If true, does not save changes to any models, but still returns the list of models that would change.
     :schema_regex: If set, filters schemas to generate using regex.
     :return: A list of model files that were changed.
     """
-
     unit_to_iri: dict[str, str] = {}
     with backup(GENERATED_SHARED_PATHS, restore=dry_run):
-        os.chdir(os.path.join(root_dir, SCHEMA_DIR_PATH))
```
nathan-stender, line 114: Note to reviewer: I realized that with SCHEMA_DIR_PATH being an absolute path from path_util, there is no need for root_dir/chdir

stephenworlow (approved):

PR #452, opened 2024-06-06
Improve luminex exponent parser

alejandro-salgado:
add luminex mark to tests
change hardcoded number of header columns with dynamic reading
set quotes as optional in regex for luminex parsing


tests/parsers/luminex_xponent/luminex_xponent_structure_test.py
```diff
@@ -89,6 +90,7 @@ def test_create_header() -> None:
     )
 
 
+@pytest.mark.luminex
```
slopez-b, line 93: Please add this mark to the `tool.pytest.ini_options` `markers` list in the `project.toml` so it does not complain when running the tests.

slopez-b (changes requested):
LGTM, just a minimal change requested

tests/parsers/luminex_xponent/luminex_xponent_structure_test.py
nathan-stender: Note: I'm not really opposed to adding these marks, however, given that you can already run tests just one parser by using `hatch run test tests/parsers/<parser>`, I wonder what the use case for this is that specifying the folder does not solve?

alejandro-salgado: With this way of running the tests, there is no need for the pytest mark

PR #461, opened 2024-06-10
internal: Verify that there are no repeated IDs that we expect to be unique in all content tests

nathan-stender:


src/allotropy/testing/utils.py
```diff
@@ -32,12 +33,57 @@ def _replace_asm_converter_name_and_version(allotrope_dict: DictType) -> DictTyp
     return new_dict
 
 
+# List of keys in ASM with "identifier" that do not need to be unique.
+# Currently, only "measurement identifier" and "calculated data document" identifier should be unique.
+# However, is is better to positive exceptions, so we don't accidentially miss a newly added unique identifier.
```
stephenworlow, line 38:
```suggestion
# However, it is better to have positive exceptions, so we don't accidentally miss a newly added unique identifier.
```

mdlu (approved):

PR #466, opened 2024-06-14
Redefine plate well count as optional in AppBio Quantstudio

alejandro-salgado:
- Redefine plate well count as optional in AppBio Quantstudio


slopez-b (approved):

nathan-stender (approved):

bwolfe-benchling:
@alejandro-salgado it would be good if we could to add a test and test file that contains a plate Block that has "Fast 96-Well Block" to test that we support those block types now and prevent future regressions.

alejandro-salgado:
@bwolfe-benchling that is a good idea. I added some test to cover those cases.

PR #472, opened 2024-06-21
chore: Add James to authors!

nathan-stender:


james-leinas (approved):
Thanks @nathan-stender!

PR #485, opened 2024-06-27
feat: allow no target DNA reference in AppBio Quantstudio design and anlysis

alejandro-salgado:
- Allow no target DNA reference in AppBio Quantstudio design and anlysis


src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -216,22 +216,23 @@ def build_delta_ct_mean(
     else:
         return None
 
-    if r_adj_eq_ct_mean_ref := build_adj_eq_ct_mean(view_data, sample, r_target):
-        data_sources.append(
-            DataSource(
-                feature="adjusted equivalent ct mean",
-                reference=r_adj_eq_ct_mean_ref,
+    if r_target is not None:
```
nathan-stender, line 219: Nit: do 

```
if r_target is None:
    return None

...
```

This will let you unindent the larger body

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -243,55 +244,58 @@ def build_delta_ct_mean(
...
+    data_sources = []
 
-    r_ct_sd_ref = build_ct_sd(view_data, sample, r_target)
-    if r_ct_sd_ref is None:
+    if ct_sd_ref := build_ct_sd(view_data, sample, target):
+        data_sources.append(DataSource(feature="ct sd", reference=ct_sd_ref))
+    else:
         return None
 
+    if r_target is not None:
```
nathan-stender, line 260: If r_target is None, I think you want to return None here, right?

If so, suggest to move `if r_target is None -> return None` to the top.

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -243,55 +244,58 @@ def build_delta_ct_mean(
...
+    data_sources = []
+    if ct_se_ref := build_ct_se(view_data, sample, target):
+        data_sources.append(DataSource(feature="ct se", reference=ct_se_ref))
+    else:
         return None
 
-    r_ct_se_ref = build_ct_se(view_data, sample, r_target)
-    if r_ct_se_ref is None:
-        return None
+    if r_target is not None:
```
nathan-stender, line 288: Same suggestion as above

nathan-stender (changes requested):

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
alejandro-salgado: the objective of this function is basically to construct the data_source list

the first value is added before this piece of code in lines 203 or 210

the second value is added or not depending on r_target. 
But if r_target is none I still have the first result. 

If there is a value in r_target I have to call build_adj_eq_ct_mean or build_eq_ct_mean trying to construct the related result, only when both functions fail I have to return the None value of line 235.

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
alejandro-salgado: is the same situation as above, if r_target is none I just have to ignore the invocation of build_ct_sd with r_target
because data_sources contains a value that I can't ignore.

nathan-stender (approved):
Ah, I see 👍

PR #489, opened 2024-06-28
feat: remove stage number from AppBio Quantstudio design and analysis

alejandro-salgado:
The QuantStudio Design & Analysis structure.py is expecting a Stage Number column within the MultiCompontent tab - which is not always present.

The Stage Number should not be necessary for structuring the amplification data cubes. Only `Well Position`, `Cycle Number`, and the associated reporter dye are the only information that should be necessary to structure the data.


tests/parsers/appbio_quantstudio_designandanalysis/testdata/appbio_quantstudio_designandanalysis_QS7Pro_Genotyping_example07.json
```diff
@@ -349,11 +350,13 @@
                                             37,
                                             38,
                                             39,
-                                            40
+                                            40,
+                                            1
```
nathan-stender, line 354: This result seems unlikely to be desirable. Are we sure there isn't some other filter we should be applying _instead_ of stage number? 

E.g. the description mentions cycle number and reporter dye setting, perhaps one of those?

nathan-stender (changes requested):

tests/parsers/appbio_quantstudio_designandanalysis/testdata/appbio_quantstudio_designandanalysis_QS7Pro_Genotyping_example07.json
alejandro-salgado: Yes, the result is weird. However James and Joe told me this is the correct change.

nathan-stender: Thank you @alejandro-salgado

@james-leinas will follow your lead on approving

james-leinas: @nathan-stender, @alejandro-salgado and I discussed async about handling stage number as filter for cases when it is present in the example file. Other cases we identified will not contain stage number and we will record all the values in the datacube for cycle number/reporter dye setting pair. This at least will help to filter when we have a basis to do so, otherwise we take the information as given to us in the file.

nathan-stender: Sorry not clear if this means there will be changes or if this is ready to do?

james-leinas: Sorry for confusion, this will mean some more changes.

PR #490, opened 2024-06-28
feat: Standardize use of "N/A" for strings where a non-applicable value is necessary

joshua-benchling:
The adapters should have a standard behavior of using "N/A" for string fields where returning the field in the JSON is expected but the value of that field corresponds to 'not applicable', 'not available', 'not accessible' or 'no answer'.


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -39,7 +39,7 @@
     try_str_from_series_or_none,
 )
 
-UNDEFINED_SAMPLE_NAME = "N/A"
+NOT_APPLICABLE = "N/A"
```
joshua-benchling, line 42: I also considered here exctracing this variable to the allotrope constants, what do you guys thing about that?

slopez-b: I think that's a good idea for consistency.

nathan-stender: I agree 👍 as part of this change, i think it makes sense to replace all "N/A" literals with a constant.

Thanks for doing this!

nathan-stender (approved):

PR #498, opened 2024-07-09
fix: Support non-numeric emission values

dchou-benchling:
For non-numeric emission values, the following fields will be set to NaN: excitation_wavelength_setting, detector_wavelength_setting, detector_bandwidth_setting


tests/parsers/agilent_gen5/testdata/luminescence/luminescence_emission_text_val.txt
```diff
@@ -0,0 +1,126 @@
...
+	Integration Time: 0:00.30 (MM:SS.ss)
+	Filter Set 1
+	    Emission: Hole
+	    Optics: Top,  Gain: 200
+	Read Speed: Normal,  Delay: 100 msec
+	Extended Dynamic Range
+	Read Height: 1 mm
+
+Layout
+	1	2	3	4	5	6	7	8	9	10	11	12	13	14	15	16	17	18	19	20	21	22	23	24
```
nathan-stender, line 34: Is there any amount of this that can be trimmed down while still exercising the new code, e.g. reducing the layout data here?

I know some instruments/cases have hard-coded expectations about layout, but if there's any possibility of trimming it down that would be good.

For new test cases (and eventually going back over existing stuff) I want us to think about minimal test cases to cover functionality, because the ASM output gets so big.

nathan-stender (dismissed):
Good looks good - one request about test data, but not blocking.

tests/parsers/agilent_gen5/testdata/luminescence/luminescence_emission_text_val.txt
dchou-benchling: Yep totally reasonable, honestly we don't even really need this new file because it turns out we have existing files that this test case falls under. @james-leinas do you have thoughts? I can trim this one down or we can consider it covered by the "Full Light" files

james-leinas: @dchou-benchling since we have the other files that the test case falls under, I'm ok with leaving this example file out of the PR.

slopez-b: @dchou-benchling If you want to test the `excitation_wavelength_setting`, `detector_wavelength_setting` and `detector_bandwidth_setting` for non-numeric  emission values. You can add tests to the `agilent_gen5_structure_test.py` and avoid having to add more example files (you can use `test_create_read_data_luminescence_full_light()` as reference)

dchou-benchling:
@nathan-stender @slopez-b (or firechief @mdlu ?) I am ready for another review, thanks!

mdlu (dismissed):

dchou-benchling:
@mdlu I had to resolve a merge conflict due to another PR that impacted the test file in mine, sorry!

nathan-stender (approved):

PR #509, opened 2024-07-12
refactor: Implement structure for nanodrop parser

nathan-stender:
Adds a structure class to nanodrop eight parser.

The main goal/benefit of this is removing accessing pandas DataFrame by row/i, increasing code clarity

It also removes a hacky way to add uuids to measurements: previous it was adding a pair of "uuid" rows to the dataframe rows. Now, it creates uuids when needed to create a document.

Finally, it removes some (though not all) of the special case handling for different wavelengths: collapsing special functions into general ones parameterized on wavelength. Some special case handling is still required, since there is some business logic around `mass_concentration`


pyproject.toml
```diff
@@ -166,7 +166,7 @@ extra-dependencies = [
 ]
 
 [tool.hatch.envs.test.scripts]
-test = "pytest -n auto {args:tests}"
+test = "pytest {args:tests}"
```
nathan-stender, line 169: Note to reviewer: I discovered that `-n auto` makes running individual tests slower, so removing by default (can still be passed in)

tests/parsers/thermo_fisher_nanodrop_eight/testdata/Thermo_NanoDrop_Eight_example01.json
```diff
@@ -7,7 +7,7 @@
                     "measurement time": "2020-05-26T15:36:00+00:00",
                     "measurement document": [
                         {
-                            "measurement identifier": "THERMO_FISHER_NANODROP_EIGHT_TEST_ID_7",
+                            "measurement identifier": "THERMO_FISHER_NANODROP_EIGHT_TEST_ID_0",
```
nathan-stender, line 10: Note to reviewer: the diffs in ids here are because we now only generate a uuid when we're going to make a measurement document, whereas previously we generated uuids for every row in the dataframe, and then threw some out (note how the test ids now start at 0)

src/allotropy/parsers/utils/iterables.py
```diff
@@ -0,0 +1,13 @@
+from collections.abc import Callable
+from typing import TypeVar
+
+T = TypeVar("T")
+S = TypeVar("S")
+
+
+def get_first(func: Callable[[T], S | None], args: list[T]) -> S | None:
```
stephenworlow, line 8: nit: it wasn't immediately obvious to me that this was "get first element that satisfies condition" without reading the code

stephenworlow (approved):

stephenworlow (approved):

PR #516, opened 2024-07-12
chore: Add script to bump PR version automatically

nathan-stender:
Uses can now automatically bump allotropy version and push a PR via: `hatch run scripts:update-version`

Example PR: https://github.com/Benchling-Open-Source/allotropy/pull/515


mdlu (commented):
could you link an example to a PR that gets autogenerated? just for reference

also could you update the README to explain what this script is and how to use it?

stephenworlow (approved):

PR #517, opened 2024-07-12
refactor: Add structure class for nucleoview parser

nathan-stender:
Encourage code reuse and clarity


pyproject.toml
```diff
@@ -168,7 +168,7 @@ extra-dependencies = [
 ]
 
 [tool.hatch.envs.test.scripts]
-test = "pytest -n auto {args:tests}"
+test = "pytest {args:tests}"
```
mdlu, line 171: doesn't this undo the speedup change you made recently?

nathan-stender: Oh sorry - I commented on another PR where I removed this 😓 

I found that this makes individual tests slower (e.g. running just this parser goes from 1s -> 8s) 

So, I am removing this and announcing to the team that it _can be_ used if they want by adding `-n auto`

nathan-stender: Upon more time with this, I decided to add it back. Making short tests take a couple more seconds, while making long tests reliably fast seems like a good tradeoff.

stephenworlow (approved):

PR #520, opened 2024-07-12
feat: redefine stage number as optional in AppBio Quantstudio design and analysis

alejandro-salgado:
- Redefine stage number as optional in AppBio Quantstudio design and analysis

This PR is another approach of https://github.com/Benchling-Open-Source/allotropy/pull/489
The idea is not to ignore the stage number entirely but to define it as optional.


nathan-stender (dismissed):
🎉

slopez-b (dismissed):

slopez-b (approved):

PR #526, opened 2024-07-13
refactor: Add Mapper to map from structured data to cell counter ASM model 

nathan-stender:

Apply this pattern to the Beckman parsers.


joshua-benchling (approved):

src/allotropy/parsers/beckman_vi_cell_blu/vi_cell_blu_parser.py
```diff
@@ -1,98 +1,19 @@
...
-    VICELL_BLU_SOFTWARE_NAME,
-)
 from allotropy.parsers.beckman_vi_cell_blu.vi_cell_blu_reader import ViCellBluReader
+from allotropy.parsers.beckman_vi_cell_blu.vi_cell_blu_structure import create_data
 from allotropy.parsers.release_state import ReleaseState
-from allotropy.parsers.utils.uuids import random_uuid_str
 from allotropy.parsers.vendor_parser import VendorParser
 
 
-class SampleProperty(Enum):
```
mdlu, line 47: double checking, do you think there's value in keeping the enum to be able to view properties at a glance? or does this basically just mirror the ASM so we don't really need it?

also confirming the original value of the Enum was more to be able to use the `data_types` but we don't need it anymore since we pass directly into the ASM dataclass?

(for the record, I am happy to remove it but just making sure it no longer provides value)

src/allotropy/parsers/beckman_vi_cell_xr/constants.py
```diff
@@ -6,9 +6,18 @@ class XrVersion(str, Enum):
...
 
-DATE_HEADER = {
-    XrVersion._2_06: "Sample date/time",
-    XrVersion._2_04: "Sample date",
+DATE_HEADER = "Sample date"
+
+HEADINGS_TO_PARSER_HEADINGS = {
+    "RunDate": DATE_HEADER,
+    "Sample date/time": DATE_HEADER,
+    # NOTE: data here is a typo in the Beckman software.
```
mdlu, line 14: no way 😅

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_parser.py
```diff
@@ -1,72 +1,23 @@
...
 from allotropy.parsers.beckman_vi_cell_xr.vi_cell_xr_reader import ViCellXRReader
+from allotropy.parsers.beckman_vi_cell_xr.vi_cell_xr_structure import create_data
 from allotropy.parsers.beckman_vi_cell_xr.vi_cell_xr_txt_reader import ViCellXRTXTReader
 from allotropy.parsers.release_state import ReleaseState
-from allotropy.parsers.utils.uuids import random_uuid_str
 from allotropy.parsers.vendor_parser import VendorParser
 from allotropy.types import IOType
 
 
-class SampleProperty(Enum):
```
mdlu, line 50: same question as for the above Enum

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_txt_reader.py
```diff
@@ -1,50 +1,22 @@
...
-    "Viability (%)",
-    "Viable cells/ml (x10^6)",
-]
-
 
 class ViCellXRTXTReader:
     def __init__(self, contents: NamedFileContents) -> None:
         self.lines = read_to_lines(contents)
         self.file_info = self._get_file_info()
-        self.file_version = self._get_file_version()
```
mdlu, line 47: double checking, we don't need this anymore because we can handle both `Sample date/time` and `Sample date` now and it wasn't used otherwise?

src/allotropy/allotrope/schema_mappers/adm/cell_counting/benchling/_2023/_11/cell_counting.py
```diff
@@ -0,0 +1,200 @@
...
+    measurements: list[Measurement]
+    _analyst: str | None = None
+
+    @property
+    def analyst(self) -> str | None:
+        if self._analyst is not None:
+            return self._analyst
+        if (
+            self.measurements
+            and len({m.analyst for m in self.measurements}) == 1
```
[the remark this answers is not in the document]
nathan-stender: Yup. I should (and will) add a comment here. The idea here is that if there is a single measurement time for all docs, then put it in the aggregate document, but otherwise don't.

src/allotropy/allotrope/schema_mappers/adm/cell_counting/benchling/_2023/_11/cell_counting.py
```diff
@@ -0,0 +1,200 @@
...
+            measurement_aggregate_document=MeasurementAggregateDocument(
+                measurement_document=[
+                    self._get_measurement_document(measurement, metadata)
+                    for measurement in measurement_group.measurements
+                ]
+            ),
+        )
+
+    def _get_measurement_document(
+        self, row: Measurement, metadata: Metadata
```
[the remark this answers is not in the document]
nathan-stender: Oh yeah forgot to change this.

src/allotropy/parsers/beckman_vi_cell_blu/vi_cell_blu_parser.py
nathan-stender: Yeah - IMHO the body of the `Measurement` creation basically lists out these values.

It would be one thing if adding to this class automatically mapped to the enum (and that may be coming :D) but at this time, it's just another place to make a change.

And yes - the data_type is no longer needed because this is handled by the mapper.

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_txt_reader.py
nathan-stender: Yup exactly 👍

mdlu (approved):

PR #527, opened 2024-07-14
refactor: Add Mapper to map from structured data to plate reader ASM model

nathan-stender:

Apply this pattern to a few easy parsers


tests/parsers/agilent_gen5/testdata/fluorescence/alphalisa_test_2.json
```diff
@@ -26,7 +26,7 @@
                                 ]
                             },
                             "sample document": {
-                                "sample identifier": "",
+                                "sample identifier": "Plate 1 A1",
```
nathan-stender, line 29: Note to reviewer: this is a bug fix, this should have been populated.

tests/parsers/agilent_gen5_image/agilent_gen5_image_structure_test.py
```diff
@@ -4,112 +4,14 @@
...
     InstrumentSettings,
-    LayoutData,
     ReadData,
     ReadSection,
 )
 from allotropy.parsers.agilent_gen5_image.constants import DetectionType
 from allotropy.parsers.lines_reader import LinesReader
 
 
-@pytest.mark.short
```
nathan-stender, line 17: Note to reviewer: these tests are deleted because gen_5_image structure re-uses identical functions from gen_5 which are already tested

src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
```diff
@@ -0,0 +1,524 @@
...
+                    data
+                ),
+            ),
+            field_asm_manifest=self.MANIFEST,
+        )
+
+    def _get_measurement_document_item(
+        self, measurement: Measurement, metadata: Metadata
+    ) -> MeasurementDocumentItems:
+        match measurement.type_:
```
mdlu, line 268: I guess we can't use this anymore 😭

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -381,200 +380,251 @@ def _get_filter_sets(
...
-        return sources or measurements
+def _create_measurement(
+    measurement: MeasurementData,
+    well_position: str,
+    header_data: HeaderData,
+    read_data: ReadData,
+    sample_identifier: str | None,
+    actual_temperature: float | None,
+) -> Measurement:
+    match read_data.read_mode:
```
mdlu, line 540: same here 😔

tests/parsers/agilent_gen5/agilent_gen5_structure_test.py
```diff
@@ -336,14 +337,37 @@ def test_create_layout_data_with_name_rows() -> None:
         "Layout",
         "\t1\t2\t3",
         "A\tSPL1\tSPL9\tSPL17\tWell ID",
-        "\tName_A1\tName_A2\tName_A3\tName",
```
mdlu, line 339: why did this go away?

src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
nathan-stender: Noooooooooooooo 😭

tests/parsers/agilent_gen5/agilent_gen5_structure_test.py
nathan-stender: Oh yeah should have left a comment. So the logic here is that the `Name` column overrides the `Well ID` comment for identifiers, if present. I'm doing a little more magic now, so I'm testing here that it works on a per-column basis. Note that "A2" below now uses the Well ID because the name is missing for that column

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
nathan-stender: 😭

mdlu (approved):

PR #532, opened 2024-07-16
refactor: Add Mapper to map from structured data to spectroscopy ASM model

nathan-stender:

Apply this pattern to thermo nanodrop eight


tests/parsers/thermo_fisher_nanodrop_eight/testdata/Thermo_NanoDrop_Eight_example01.json
```diff
@@ -48,7 +48,7 @@
                     "measurement time": "2020-05-26T15:36:00+00:00",
                     "measurement document": [
                         {
-                            "measurement identifier": "THERMO_FISHER_NANODROP_EIGHT_TEST_ID_1",
```
nathan-stender, line 51: Note to reviewer: these changes are because calculated data document uuids are now calculated at the same time as the measurment, so the order changes. This does not matter for real data that actually uses uuids

stephenworlow (approved):

PR #534, opened 2024-07-16
refactor: Refactor series utils and break up values util

nathan-stender:
Basically, this adds a wrapper for a `pd.Series` that replaces our endless util functions for asserting type safety with a cleaner API:

`try_str_from_series_or_none(data, "key")` -> `data.get(str, "key")`

`try_str_from_series(data, "key")` -> `data[str, "key"]` (raises error on missing key)

`try_str_from_series(data, "key", msg="error msg")` -> `data[str, "key", "error msg"]` (raises "error msg" on missing key)

`try_str_from_series_or_default(data, "key", "default")` -> `data.get(str, "key", "default")`

`try_non_nan_str_from_series_or_none(data, "key")` -> `data.get(str, "key", SeriesData.NOT_NAN)`

`try_float_from_series_or_nan(data, "key")` -> `data.get(float, "key", NaN)`


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -571,50 +517,43 @@ def create_generic(data: pd.DataFrame, well_item: WellItem) -> Result:
...
-            ),
-            genotyping_determination_result=try_str_from_series_or_none(
-                target_data, "Call"
+            automatic_cycle_threshold_enabled_setting=data.get(
+                bool, "Automatic Ct Threshold", None
             ),
-            genotyping_determination_method_setting=try_float_from_series_or_none(
-                target_data, "Threshold Value"
+            automatic_baseline_determination_enabled_setting=data.get(
+                bool, "Automatic Baseline", None
```
mdlu, line 534: `, None` also not needed here and in the above line

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_structure.py
```diff
@@ -555,55 +527,48 @@ def create(
...
-            automatic_cycle_threshold_enabled_setting=try_bool_from_series_or_none(
-                target_data, "Auto Threshold"
+            cycle_threshold_result=target_data.get(float, "Cq"),
+            automatic_cycle_threshold_enabled_setting=target_data.get(
+                bool, "Auto Threshold", None
             ),
-            automatic_baseline_determination_enabled_setting=try_bool_from_series_or_none(
-                target_data, "Auto Baseline"
+            automatic_baseline_determination_enabled_setting=target_data.get(
+                bool, "Auto Baseline", None
```
mdlu, line 539: `, None` also not needed here and in the above line

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -151,30 +134,19 @@ def result(self, result: Result) -> None:
...
     @staticmethod
-    def create_genotyping(data: pd.Series[str]) -> tuple[WellItem, WellItem]:
-        identifier = try_int_from_series(data, "Well")
-
-        snp_name = try_str_from_series(
-            data,
-            "SNP Assay Name",
+    def create_genotyping(data: SeriesData) -> tuple[WellItem, WellItem]:
+        identifier = data[int, "Well"]
+        snp_name = assert_not_none(
```
mdlu, line 139: nit: feel free to disagree if this is too much abstraction but it feels like `assert_not_none` is called a lot with `data.get`, maybe we could add it as a parameter to `.get` or add a new function?

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_parser.py
```diff
@@ -91,11 +91,7 @@ def to_allotrope(self, named_file_contents: NamedFileContents) -> Model:
         lines = read_to_lines(named_file_contents)
         reader = CsvReader(lines)
         filename = named_file_contents.original_file_name
-        try:
-            return self._get_model(Data.create(reader), filename)
-        except Exception as error:
```
mdlu, line 96: do we not need to catch this anymore?

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py
```diff
@@ -55,61 +53,51 @@ class PlateInfo:
...
+            df_to_series(
+                assert_not_none(
+                    reader.pop_csv_block_as_df(),
+                    "Plate information CSV block",
+                )
+            ).replace(np.nan, None)
         )
 
-    @staticmethod
-    def get_barcode(series: pd.Series[str], plate_number: str) -> str:
```
mdlu, line 80: is this not useful anymore as a standalone function? I guess it's only called once?

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py
```diff
@@ -554,38 +542,29 @@ def create(reader: CsvReader) -> Labels:
...
+        series = SeriesData(df_to_series(data.T).replace(np.nan, None))
         filters = create_filters(reader)
         filter_position_map = {
             "Bottom": ScanPositionSettingPlateReader.bottom_scan_position__plate_reader_,
             "Top": ScanPositionSettingPlateReader.top_scan_position__plate_reader_,
         }
 
         return Labels(
-            label=series.index[0],
+            label=series.series.index[0],
```
mdlu, line 553: nit: `series.series` makes me kind of sad haha maybe `series_data.series`?

src/allotropy/parsers/utils/pandas.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            # bool needs special handling to convert
+            if type_ is bool:
+                raw_value = "true" if str_to_bool(raw_value) else ""
+            value = None if raw_value is None else type_(raw_value)
+        except ValueError:
+            value = None
+        return default if value is None else value
+
+    # TODO(nstender): I can't figure out how to integrate these yet, leaving as "try_..."
```
mdlu, line 105: I am definitely missing something but couldn't these be handled using an extra param plus some logic in the `ValueError` block in the above `get`? or have a flag for `.get` to be able to raise `ValueError` and then another function that wraps this function and sets that flag for additional handling

mdlu (changes requested):

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
```diff
@@ -506,63 +459,56 @@ def create_genotyping(data: pd.DataFrame, well_item: WellItem) -> Result:
...
             cycle_threshold_result=try_float_or_none(str(cycle_threshold_result)),
-            automatic_cycle_threshold_enabled_setting=try_bool_from_series_or_none(
-                target_data, f"{allele} Automatic Ct Threshold"
+            automatic_cycle_threshold_enabled_setting=data.get(
+                bool, f"{allele} Automatic Ct Threshold", None
             ),
-            automatic_baseline_determination_enabled_setting=try_bool_from_series_or_none(
-                target_data, f"{allele} Automatic Baseline"
+            automatic_baseline_determination_enabled_setting=data.get(
+                bool, f"{allele} Automatic Baseline", None
```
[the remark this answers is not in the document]
nathan-stender: Good eye - I guess my regex wasn't perfect 👀

src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_structure.py
nathan-stender: So - the issue here is that if you want to assert, you should just use the index operator.

But - these particular cases want to pass a custom error message. That said, I realized I can still support the custom error message in the index operator, I was just too tired to think of it last night 😓

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_parser.py
nathan-stender: Sorry should have left a reviewer comment.

This was a bad pattern. They were catching all errors and returning a generic error instead.

The pattern we want is to catch errors where they happen and return an `AllotropeConversionError` with a _helpful_ error message.

Removing this may let some uncaught errors through - but that's how we find and address those uncaught errors.

src/allotropy/parsers/perkin_elmer_envision/perkin_elmer_envision_structure.py
nathan-stender: Yeah, it's called twice, but it's smaller now, and, I plan to dedup the two call sites in a followup.

src/allotropy/parsers/utils/pandas.py
nathan-stender: For `non_nan` - it's more philosophical - bool flags in function signatures are discouraged. I might just go with it, but I deferred until later.

For `try_float_or_nan` - the problem is a typing one, it should just be:

`self.get(float, key, InvalidJsonFloat.NaN)`

but typing of the `get` function does not allow for returning `InvalidJsonFloat` when `T = float`. I'm sure it's doable, but haven't figured it out yet.

nathan-stender: Aaand fixed. Thank you for pushing me 🙏 😄

src/allotropy/parsers/utils/values.py
```diff
@@ -16,7 +16,7 @@
 PrimitiveValue = str | int | float
 
 
-def str_to_bool(value: Any) -> bool:
+def str_to_bool(value: str) -> bool:
     return str(value).lower() in ("yes", "y", "true", "t", "1")
```
mdlu, line 20: nit: don't need to cast `value` anymore in that case?

src/allotropy/parsers/utils/pandas.py
```diff
@@ -53,50 +53,58 @@ def assert_not_empty_df(df: pd.DataFrame, msg: str) -> pd.DataFrame:
...
 
 
 class SeriesData:
     def __init__(self, series: pd.Series[Any]) -> None:
         self.series = series
 
-    def __getitem__(
-        self, type_and_key: tuple[Callable[..., T], str | Iterable[str]]
-    ) -> T:
+    def __getitem__(self, type_and_key: TypeAndKey[T] | TypeAndKeyAndMsg[T]) -> T:
```
mdlu, line 81: I realized that this would really benefit from a docstring explaining how to use this (what the params are, and that this implements the `series[]` method of get)

src/allotropy/parsers/utils/pandas.py
```diff
@@ -53,50 +53,58 @@ def assert_not_empty_df(df: pd.DataFrame, msg: str) -> pd.DataFrame:
...
+        type_: Type_[T],
+        key: KeyOrKeys,
         default: Literal[None] = None,
     ) -> T | None:
         ...
 
     # This overload tells typing that if default matches T, get will return T
     @overload
-    def get(self, type_: Callable[..., T], key: Iterable[str], default: T) -> T:
+    def get(self, type_: Type_[T], key: KeyOrKeys, default: T) -> T:
```
mdlu: same thing here, as above

mdlu (changes requested):

src/allotropy/parsers/utils/values.py
nathan-stender: Very true

mdlu (approved):
🚀

PR #538, opened 2024-07-18
fix: Allow ASM schema models to further combine model classes

nathan-stender:
The `solution-analyzer` schema had too many different possible combinations of required keys for some models, leading to some >60 versions of possible measurement document. 

Relax schema model generator class combination logic further to remove this explosion of classes.

Note that as with existing relaxations - this means it's possible to create a version of the dataclass that does not satisfy the schema (now even more for affected schemas)

However, it's still not possible to write a test with an invalid schema or export an invalid schema from allotropy, since we validate against the schema after conversion.


src/allotropy/allotrope/schema_parser/model_class_editor.py
```diff
@@ -373,8 +363,16 @@ def merge_similar(self, other: DataClassLines) -> DataClassLines:
...
                 self.fields[field_name] = other.fields[field_name]
+                self.fields[field_name].field_types.add("None")
+                self.fields[field_name].default_value = "None"
                 self.field_name_order.append(field_name)
 
+        # For fields in this class not in the other, mark as optional
+        for field_name in self.fields:
+            if field_name not in other.fields:
+                self.fields[field_name].field_types.add("None")
+                self.fields[field_name].default_value = "None"
```
stephenworlow, line 374: Nit: Mostly because the code is kinda ugly, wanna abstract this and above?

stephenworlow (dismissed):

src/allotropy/allotrope/schema_parser/model_class_editor.py
nathan-stender: Done

stephenworlow (approved):

PR #539, opened 2024-07-19
feat: Speed up excel reading by using calamine engine where possible

nathan-stender:
This cuts a solid amount of time off of larger excel file test cases. 

E.g. in a test file that was timing out a web API because it took ~40s to convert, it now takes ~20s


tests/parsers/beckman_vi_cell_xr/testdata/v2.06/style_fill_incorrect.json
```diff
@@ -71,7 +71,7 @@
             "equipment serial number": "AL42244"
         },
         "data system document": {
-            "file name": "style_fill_error.xlsx",
+            "file name": "style_fill_incorrect.xlsx",
```
nathan-stender, line 74: NOTE to reviewer: I had to rename this file because the "error" in the name was causing the auto-test discovery to skip it.

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
```diff
@@ -1,32 +1,59 @@
...
     MODEL_RE,
     XrVersion,
 )
+from allotropy.parsers.lines_reader import read_to_lines
 from allotropy.parsers.utils.pandas import SeriesData
 from allotropy.parsers.utils.values import assert_value_from_df
-from allotropy.types import IOType
+
+
+@dataclass
```
nathan-stender, line 24: Note to reviewer: I added this little refactor because I noticed that the `structure` class isn't actually using the reader except to access these two pieces of data, so just return them.

stephenworlow (dismissed):

stephenworlow (approved):

PR #545, opened 2024-07-23
feat: Add BMG MARS parser

keithmacaulay:


src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
```diff
@@ -0,0 +1,270 @@
...
+        read_type: ReadType,
+    ) -> DeviceControlAggregateDocument:
+
+        if read_type == ReadType.ABSORBANCE:
+            return UltravioletAbsorbancePointDetectionDeviceControlAggregateDocument(
+                device_control_document=[
+                    UltravioletAbsorbancePointDetectionDeviceControlDocumentItem(
+                        device_type="absorbance detector",
+                        detection_type="absorbance",
+                        detector_wavelength_setting=safe_value(
```
nlangowitz, line 158: A bit surprised by the checks here -- seems like `wavelength` must be set, so the `if wavelength.wavelength` check would always pass unless wavelength is 0 or an empty string, are those conditions we'd expect to see in a data file? If not, seems like this could just be `TQuantityValueNanometer(wavelength.wavelength)`, and if so, I think it'd be better to check for that when creating the `Wavelength` object.

src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
```diff
@@ -0,0 +1,270 @@
...
+            return FluorescencePointDetectionDeviceControlAggregateDocument(
+                device_control_document=[
+                    FluorescencePointDetectionDeviceControlDocumentItem(
+                        device_type="fluorescence detector",
+                        detection_type="fluorescence",
+                        detector_wavelength_setting=safe_value(
+                            TQuantityValueNanometer,
+                            wavelength.wavelength if wavelength.wavelength else None,
+                        ),
+                        excitation_wavelength_setting=safe_value(
```
nlangowitz, line 176: The `safe_value` check makes sense here, but feels overcomplicated since we have the if/else both here and in the function, why not just `TQuantityValueNanometer(wavelength.ex_wavelength) if wavelength.ex_wavelength else None`

src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
```diff
@@ -0,0 +1,270 @@
...
+        read_type = assert_not_none(
+            re.search(RE_READ_TYPE, "\n".join(lines), flags=re.MULTILINE),
+            msg="Read type not found.",
+        ).group(0)
+        patterns = {
+            "Absorbance": ReadType.ABSORBANCE,
+            "Fluorescence": ReadType.FLUORESCENCE,
+        }
+
+        for key in patterns:
```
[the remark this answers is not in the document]
keithmacaulay: We don't have examples of this, so removing.

src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
keithmacaulay: Simplified detector_wavelength_setting for UV and Fluorescence

src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
keithmacaulay: Simplified detector_wavelength_setting for UV and Fluorescence

src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
```diff
@@ -133,15 +134,13 @@ def _get_read_type(self, lines: list[str]) -> ReadType:
...
-        patterns = {
-            "Absorbance": ReadType.ABSORBANCE,
-            "Fluorescence": ReadType.FLUORESCENCE,
-        }
 
-        for key in patterns:
-            if key in read_type:
-                read_type_key = patterns[key]
-        return read_type_key
+        if "Absorbance" in read_type:
```
rayaankha-tara, line 138: Just confirming that if these are somehow both in the read_type that "Absorbance" takes precedence? Sorry this may just be a nonsensical scenario

rayaankha-tara (dismissed):
approving changes since @nlangowitz 's review with minor question

src/allotropy/parsers/bmg_mars/bmg_mars_parser.py
keithmacaulay: Updated to explicitly exclude the scenario of >1 read_type.

rayaankha-tara (approved):

PR #547, opened 2024-07-25
fix: agilent tapestation ASM fixes

slopez-b:
- For the Agilent Tapestation adapter, the data region document was being included even when there was no region data, resulting on empty lists. This PR adds logic to not include the data region aggregate document if there is no region data.

- Additionally, removed the `devide identifier` from the `device control document` since it is already included in the `device system document` at a top level.


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Fixed
+- Fixed tapestation analysis adapter to not inlude data region documents when there is no region data.
+- Removed `devide identifier` from `device control document` in the tapestation analysis adapter.
```
nlangowitz, line 14:
```suggestion
- Removed `device identifier` from `device control document` in the tapestation analysis adapter.
```

nlangowitz (dismissed):

joshua-benchling (approved):

PR #548, opened 2024-07-26
fix: ViCell Blu adapter not recognizing columns with µ character

slopez-b:
Some column names in the instrument files include two different `µ` characters that look the same,
the `Micro Sign` (`U+00B5`) and the `Greek Small letter mu`  (`U+03BC`). This PR adds a mapper to the ViCell Blu reader that normalizes the column names and avoid errors when trying to retrieve values from those columns.


rayaankha-tara (approved):
awesome 🔥  jumping into review this so we can close out the relevant ticket

PR #552, opened 2024-08-01
feat: use solution analyzer for  novabio flex adapter

slopez-b:
Use the Solution Analyzer Schema in the NovaBio Flex adapter


src/allotropy/parsers/novabio_flex2/novabio_flex2_parser.py
```diff
@@ -24,51 +45,143 @@ def release_state(self) -> ReleaseState:
...
-            measurement_time=self._get_date_time(sample.measurement_time),
-            analyte_aggregate_document=AnalyteAggregateDocument(
-                analyte_document=[
-                    AnalyteDocumentItem(
-                        analyte_name=analyte.name,
-                        molar_concentration=analyte.molar_concentration,
+            device_control_aggregate_document=DeviceControlAggregateDocument(
+                device_control_document=[
+                    DeviceControlDocumentItem(
+                        device_type=DEVICE_TYPE, detection_type=detection_type
```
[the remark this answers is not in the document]
slopez-b: correct, this was added to differentiate the type of measurement but was not in the requirements.

joshua-benchling (approved):

src/allotropy/parsers/novabio_flex2/novabio_flex2_structure.py
```diff
@@ -65,25 +69,27 @@ def __lt__(self, other: Any) -> bool:
...
+    def create(cls, series: pd.Series[Any]) -> Sample:
         data = SeriesData(series)
-        properties: dict[str, Any] = {
-            property_name: property_dict["cls"](
-                value=data.get(float, property_dict["col_name"])
-            )
-            for property_name, property_dict in PROPERTY_MAPPINGS.items()
-            if data.get(float, property_dict["col_name"]) is not None
-        }
+        cell_density_dilution = data.get(str, "Cell Density Dilution", "")
```
rayaankha-tara, line 86: "Cell Density Dilution" Is this useful to be a constant

src/allotropy/parsers/novabio_flex2/novabio_flex2_structure.py
```diff
@@ -93,9 +99,30 @@ def create(series: pd.Series[Any]) -> Sample:
...
-            properties=properties,
+            cell_counter_properties=cls._get_properties(data, CELL_COUNTER_MAPPINGS),
+            blood_gas_properties=cls._get_properties(
+                data, BLOOD_GAS_DETECTION_MAPPINGS
+            ),
+            osmolality_properties=cls._get_properties(
+                data, OSMOLALITY_DETECTION_MAPPINGS
+            ),
+            ph_properties=cls._get_properties(data, PH_DETECTION_MAPPINGS),
+            cell_type_processing_method=data.get(str, "Cell Type"),
```
rayaankha-tara, line 110: I know this is not accessed on class properties but ditto any value to this as a constant?

rayaankha-tara (approved):
Minor/possibly ignoreable comments. One admin comment: I dont think this should be called a refactor, this would cause functional changes for allotropy even though its awesome that we are using the allotrope provided manifest

src/allotropy/parsers/novabio_flex2/novabio_flex2_structure.py
nathan-stender: I think for property access strings the answer is no. These are one-time-use and very widespread (we're always having to access columns of dataframes). It would be onerous to make a constant for every one.

src/allotropy/parsers/novabio_flex2/novabio_flex2_structure.py
nathan-stender: Ditto

PR #557, opened 2024-08-06
feat: Handle comma separated decimal in float util

nathan-stender:


src/allotropy/parsers/utils/values.py
```diff
@@ -35,18 +35,25 @@ def try_int_or_none(value: str | None) -> int | None:
...
+    # an unexpected error, e.g. one thousand represented as 1,000 would get converted to 1.0
+    # However, numbers are not usually represented like this in scientific output (we have no example of it)
+    return float(value.replace(",", "."))
+
+
 def try_float(value: str, value_name: str) -> float:
     assert_not_none(value, value_name)
     try:
-        return float(value)
+        return _try_float(str(value))
```
mdlu, line 48: nit: if `value` is already a `str` do we still need to cast?

mdlu (dismissed):

src/allotropy/parsers/utils/values.py
nathan-stender: Jeez, always counting me out on my laziness :P 

Very fair. There were some untyped calls to this functions that weren't necessarily string (discovered by failing tests). I hunted them down and fixed them instead.

mdlu (approved):

PR #558, opened 2024-08-06
refactor: Remove converting measurement calc docs to global calc docs

nathan-stender:
This removes a mistaken approach of pulling all calculated documents from all measurement documents into the global calculated data documents list. This is wrong because some ASM documents report calculated data documents inside of measurement documents.

The location of the calculated data documents within the data classes should signal where they should be placed within the ASM document.

This was done in the `plate reader` mapper because one parser (unchained labs) was taking this approach, of putting all calculated data documents in the global document, even if they correspond to measurement rows. I'm not sure why we are putting measurement-specific calculated data documents in the global calculated data document, and I'm not changing it here (left a TODO comment to investigate).

Instead, the offending parser is changed to move the calculated documents to the global level in the data class. 

Note that this is classified as a refactor only, because it does not change any output of any test cases.

### DIFF
--- a/src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
+++ b/src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
@@ -26,7 +26,9 @@ from allotropy.parsers.utils.uuids import random_uuid_str
 
 
 def _create_measurement(
-    well_plate_data: SeriesData, wavelength_column: str
+    well_plate_data: SeriesData,
+    wavelength_column: str,
+    calculated_data: list[CalculatedDataItem],
 ) -> Measurement:
     if wavelength_column not in well_plate_data.series:
         msg = NO_MEASUREMENT_IN_PLATE_ERROR_MSG.format(wavelength_column)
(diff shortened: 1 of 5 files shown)


src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
```diff
@@ -36,6 +38,9 @@ def _create_measurement(
         raise AllotropeConversionError(INCORRECT_WAVELENGTH_COLUMN_FORMAT_ERROR_MSG)
 
     measurement_identifier = random_uuid_str()
+    calculated_data.extend(
```
mdlu, line 41: python is confusing... took me a while to realize that this modifies `calculated_data` 😱

mdlu (approved):

src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
nathan-stender: Yeah, pass-by-reference in python isn't really encouraged/common. 

I did it here because it's the fastest way to make this work, while I figure out how to get rid of it :P

PR #559, opened 2024-08-07
refactor: Use mapper class for qubit4 parser

nathan-stender:
Includes adding custom info document to spectroscopy mapper


mdlu (approved):
🎉

PR #563, opened 2024-08-08
fix: Cast data to string before parsing in appbio quantstudio

nathan-stender:
Without this, using the `.str` methods below will fail. 

This does not cause any issue because the code treats all values as if they were string (using other methods to cast to float where expected)


CHANGELOG.md
```diff
@@ -13,9 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add BMG MARS adapter
 
 ### Fixed
+- Handle comma as decimal place in float conversion utilities
```
stephenworlow, line 16: I think this is already below?

stephenworlow (approved):

CHANGELOG.md
nathan-stender: Gahhhh

PR #566, opened 2024-08-09
fix: Simplify Vi-cell XR text reader

nathan-stender:
The `pivot` operation in XR text file reader can have a failure if there is duplicated data, which was seen a couple times in real life usage. 

Simplify the parsing of the text file into pandas dataframe, removing the need for the pivot entirely.

### DIFF
--- a/src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
+++ b/src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
@@ -7,7 +7,8 @@ from typing import Any
 
 import pandas as pd
 
-from allotropy.allotrope.pandas_util import read_excel
+from allotropy.allotrope.pandas_util import read_csv, read_excel
+from allotropy.exceptions import AllotropyParserError
 from allotropy.named_file_contents import NamedFileContents
 from allotropy.parsers.beckman_vi_cell_xr.constants import (
     DATE_HEADER,
@@ -17,13 +18,13 @@ from allotropy.parsers.beckman_vi_cell_xr.constants import (
     XrVersion,
 )
 from allotropy.parsers.lines_reader import read_to_lines
-from allotropy.parsers.utils.pandas import SeriesData
+from allotropy.parsers.utils.pandas import df_to_series, SeriesData
 from allotropy.parsers.utils.values import assert_value_from_df
 
 
 @dataclass
 class ViCellData:
-    data: pd.DataFrame
+    data: list[SeriesData]
     file_info: SeriesData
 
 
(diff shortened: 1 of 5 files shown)


src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
```diff
@@ -78,19 +79,22 @@ def _read_data(self) -> pd.DataFrame:
             errors="coerce",
         )
         file_data = file_data.dropna(subset=DATE_HEADER)
-
-        return file_data
+        return [SeriesData(x) for _, x in file_data.iterrows()]
```
nathan-stender, line 82: Note: this change is for compatibility with text read changes, it's just converting to `SeriesData` earlier, so I don't have to bother changing the series below back to a dataframe.

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
```diff
@@ -78,19 +79,22 @@ def _read_data(self) -> pd.DataFrame:
...
 
     def _get_file_header(self, header_row: int) -> list[str]:
         """Combine the two rows that forms the header."""
-        header = self._read_excel(
-            nrows=2,
-            skiprows=header_row,
-            header=None,
-        ).fillna("")
+        header = (
```
nathan-stender, line 86: Note: this is a re-fixing of a recently fixed bug. The join below could fail if all elements weren't strings. I realized I can cast to string in pd, instead of in the list below

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
```diff
@@ -102,9 +106,15 @@ def _get_file_info(self) -> SeriesData:
         return SeriesData(info)
 
     def _get_file_version(self) -> XrVersion:
-        match = re.match(MODEL_RE, self.file_info[str, "model"], flags=re.IGNORECASE)
+        raw_model = self.file_info.get(str, "model")
+        if not raw_model:
```
nathan-stender, line 110: Note: this is a "fix" (really just getting more info) for a bug I saw where `model` key was missing. I don't see how that could happen, because we set model in the index elsewhere. But, this new error will give the info if it happens agian.

src/allotropy/parsers/beckman_vi_cell_xr/vi_cell_xr_reader.py
```diff
@@ -113,50 +123,34 @@ def _get_file_version(self) -> XrVersion:
...
-        file_data["Pivot Index"] = 1
-        file_data = file_data.pivot(index="Pivot Index", columns=0)
-
-        # rename the columns to match the existing parser that was based on xls(x) files
-        file_data.columns = pd.Index(
-            [
-                HEADINGS_TO_PARSER_HEADINGS.get(name, name)
-                for name in [x[1] for x in file_data.columns]
-            ]
+        file_data = df_to_series(
```
nathan-stender, line 138: Note: the crux of the fix. Instead of splitting the dataframe up and pivoting, use the `:` separator in the first place (with a leading space to preserve dates) and then just transpose it.

stephenworlow (approved):

PR #582, opened 2024-08-14
refactor: split structure by experiment type in AppBio Quantstudio Design and Anlysis

alejandro-salgado:
split structure by experiment type in AppBio Quantstudio Design and Anlysis

Up to this point the logic of the different experiment types is implemented using if statments
This PR intend to implement a generic structure from which each experiment type can inherit to implement their specific logic.


src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_parser.py
```diff
@@ -191,8 +191,12 @@ def get_device_control_document_item(
         return DeviceControlDocumentItem(
             device_type="qPCR",
             measurement_method_identifier=data.header.measurement_method_identifier,
-            total_cycle_number_setting=TQuantityValueNumber(
-                value=well_item.amplification_data.total_cycle_number_setting,
+            total_cycle_number_setting=(
+                TQuantityValueNumber(
+                    value=well_item.amplification_data.total_cycle_number_setting,
+                )
+                if well_item.amplification_data is not None
```
rayaankha-tara, line 198: Should this use the `quantity_or_none` helper or some thing? Apologies the name isnt coming to mind but I can find it if you dont happen to know off hand

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
```diff
@@ -147,9 +157,21 @@ def create(
...
-        amp_data = contents.get_non_empty_sheet("Amplification Data")
-        melt_curve_data = contents.get_non_empty_sheet_or_none("Melt Curve Raw")
+        amp_data = cls.get_amplification_data_sheet(contents)
+        amplification_data = (
+            AmplificationData.create(amp_data, identifier, target_dna_description)
+            if amp_data is not None
+            else None
+        )
 
+        melt_data = contents.get_non_empty_sheet_or_none("Melt Curve Raw")
```
rayaankha-tara, line 167: General comment but not specific to this pr, I wonder if a constant text string like this should in this case be a property of `MeltCurveData` or the corresponding creator class

rayaankha-tara: as in a class level prop

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
```diff
@@ -466,6 +437,14 @@ class Result:
     slope: float | None
     efficiency: float | None
 
+    @classmethod
+    def get_genotyping_determination_result(cls, _: SeriesData) -> str | None:
+        return None
```
rayaankha-tara, line 442: It seems like this can only return `None` so Im a bit confused by the str type return. Is this conforming to the method signature of some class template? Just flagging in case this class should inherit from some parent class template. Again maybe this is the existing code style so not blocking on this but definitely curious!

rayaankha-tara (changes requested):
Nice work!! To confirm, none of the other appbio test files needed to be changed to work with this code change? ie the asm `.json` test files stayed the same

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_parser.py
alejandro-salgado: I can't use quantity_or_none because the function expects a value that can be none. 
The problem is the nullable value is not total_cycle_number_setting but amplification_data.

So if I do something like 
**quantity_or_none(TQuantityValueNumber, well_item.amplification_data.total_cycle_number_setting)**
and amplification_data is none I will get an error saying None value has no attribute total_cycle_number_setting

So I have to check if amplification is none anyway

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
alejandro-salgado: it is a good idea.

I'll implement for all sheets names in a separated PR

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
alejandro-salgado: This function is design to be overwritten by the experiment types that do have a genotyping determination result

the default behavior is to not have this value, that is why the function is just a return None

alejandro-salgado:
The idea of this PR is just to change the structure of the adapter, not its functionality, so there is no need to update the tests

The new json file comes from the PR https://github.com/Benchling-Open-Source/allotropy/pull/584 in which a new experiment type is added to the structure

src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_parser.py
rayaankha-tara: Ah got it, thanks for the clarification

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
rayaankha-tara: 💯

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
rayaankha-tara: Ahhh this is the base class my bad

rayaankha-tara (approved):

PR #583, opened 2024-08-15
refactor: Add schema mapper for light-obscuration

nathan-stender:


src/allotropy/parsers/beckman_pharmspec/constants.py
```diff
@@ -1 +1,11 @@
 PHARMSPEC_SOFTWARE_NAME = "PharmSpec"
+
+# This map is used to map the column names to units, for calculated data items.
```
mdlu, line 3: this *feels* like something that should be inferable from the types (like `TQuantityValueMicrometer` should always be `μm`) but I imagine it is probably not easy to do

mdlu (approved):

src/allotropy/parsers/beckman_pharmspec/constants.py
nathan-stender: The issue is we don't know about those things at this level, we only have those when mapping to ASM.

Note that the `processed data` in this mapper do not specify unit, because we have the unit class when mapping to ASM

But, the `calculated data` docs do not specify a unit class (they are general purpose) so we need to specify the unit here.

I actually do think I might make something in the mapper to infer unit from the property name (by matching against the processed data doc) in a follow up, because typing is made at me for specifying the unit anyway (see TODO in the mapper)

PR #584, opened 2024-08-15
feat: add primary analysis experiment type to AppBio Quantstudio Design and Analysis

alejandro-salgado:
- redefine amplification data as optional in AppBio Quantstudio Design and Analysis
- add primary analysis experiment type to AppBio Quantstudio Design and Analysis


src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_parser.py
```diff
@@ -191,8 +191,12 @@ def get_device_control_document_item(
         return DeviceControlDocumentItem(
             device_type="qPCR",
             measurement_method_identifier=data.header.measurement_method_identifier,
-            total_cycle_number_setting=TQuantityValueNumber(
-                value=well_item.amplification_data.total_cycle_number_setting,
+            total_cycle_number_setting=(
```
nathan-stender, line 194:
```suggestion
            total_cycle_number_setting=quantity_or_none(TQuantityValueNumber, well_item.amplification_data.total_cycle_number_setting)
```

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
```diff
@@ -132,6 +132,12 @@ def get_result_class(cls) -> type[Result]:
     def __hash__(self) -> int:
         return hash(self.identifier)
 
+    @classmethod
+    def get_amplification_data_sheet(
```
rayaankha-tara, line 136: probably a q for another time but is there a reason we dont have `contents: DesignQuantstudioContents` as a property of objects of this `WellItem` class or said another way is there a reason this is a class method? it seems like more of an object method flavour at a glance

nathan-stender (approved):

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
```diff
@@ -151,11 +157,21 @@ def create(
             f"Unable to find well position for Well '{identifier}'.",
         ]
 
-        amp_data = contents.get_non_empty_sheet("Amplification Data")
-        melt_curve_data = contents.get_non_empty_sheet_or_none("Melt Curve Raw")
+        amp_data = cls.get_amplification_data_sheet(contents)
+        amplification_data = (
+            AmplificationData.create(amp_data, identifier, target_dna_description)
+            if amp_data is not None
+            else None
```
rayaankha-tara, line 164: Just flagging in case @nathan-stender 's comment above applies here https://github.com/Benchling-Open-Source/allotropy/pull/584/files#r1718937680

rayaankha-tara: probably not since this is not a quantity 🤦🏽

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
nathan-stender: I think this smells like over-design. Why does this method exist instead of just calling `contents.get_non_empty_sheet("Amplification Data")`?

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
nathan-stender: Yup that util is just a special case for the very common check when setting TQuantityValues in ASM

src/allotropy/parsers/appbio_quantstudio_designandanalysis/structure/generic/structure.py
alejandro-salgado: Up to this point the experiment types were handled with if statements, now the idea is to have a generic structure and when an experiment type needs a different treatment it can overwrite the default behavior. That is the reason this is a classmethod and that is the reason why the method is so simple in this case.

PR #587, opened 2024-08-15
refactor: Use schema mapper for novabio flex

nathan-stender:


src/allotropy/parsers/novabio_flex2/novabio_flex2_structure.py
```diff
@@ -116,11 +120,9 @@ def _get_properties(
         cls, data: SeriesData, property_mappings: dict[str, Any]
```
slopez-b, line 120: It seems like this method is no longer necesary.

nathan-stender: Nice catch! Done!

slopez-b (approved):

mdlu (approved):
approving for new changes since Sebastian's approval

PR #588, opened 2024-08-15
fix: improve identification of numeric columns for sofmax pro calculated data

slopez-b:
Some columns in the SoftMax Pro adapter were incorrectly identified as not numeric for calculated data documents. This was due to some values being flagged as `Masked` or `Range?`. Those cases will be treated as `NaN`.

Note: in a future version of this adapter, those values will probably be reported in the `ErrrorDocument` when we drop support for `InvalidJsonFloat` in the `tQuantityValues`


CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added support for xlsx exports in Unchained Lunatic parser
+
 ### Fixed
 
+- Fix SoftMax Pro case where some columns were incorrectly identified as not numeric for calculated data documents
```
rayaankha-tara, line 16: You can probably specify the values of `Masked` and `Range`  here too? what do you think?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -134,7 +137,7 @@ def create(data: pd.DataFrame) -> GroupSampleData:
                     plate=row[str, "WellPlateName"],
                     entries=[
                         GroupDataElementEntry(
-                            name=column_name, value=row[float, column_name]
+                            name=column_name, value=row.get(float, column_name, NaN)
```
rayaankha-tara, line 140: Would this not lead to other non `Masked` `Range?` strings being cast to NaN maybe Im missing something in the logic upstream though!

rayaankha-tara (changes requested):
High level question: I was expecting the .txt raw data files for softmax to change to test this modification. I assume the existing cases were thought to be sufficient because they include "Range?" in places correct? Just confirming that they in fact test the numeric column detection logic modified here ie that they have "Range" as a value in the top cell of a numeric column

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
nathan-stender: @rayaankha-tara your reading here is correct. 

In this case, I think it's acceptable to treat all invalid values as NaN

Since this is machine generated data, we typically don't have to be as worried about weird unintentional cases.

I think this will be further improved when the error document is added, when we can put the value that caused the nan in the error document.

nathan-stender:
Agree with @rayaankha-tara that tests should change here. If we don't have a case of the error right now, one should be added.

rayaankha-tara (approved):

nathan-stender (approved):

PR #594, opened 2024-08-16
test: Refactor schema generation script so that it does not modify files while running

nathan-stender:
This:

1. Fixes a rare flake when running all tests in parallel - if code tried to read a schema that is being modified by the schema generation test, the test could fail
2. Allows running schema generation test (the current long pole of testing) in parallel, reducing the average run time of all tests (when run in parallel) from ~45s to ~35s


src/allotropy/allotrope/schema_parser/backup_manager.py
```diff
@@ -17,19 +19,32 @@ def _files_equal(path1: PathType, path2: PathType) -> bool:
...
     _path = Path(path)
-    return Path(_path.parent, f".{_path.stem}.bak{_path.suffix}")
+    return Path(_path.parent, f".{_path.stem}.{random_uuid_str()}.bak{_path.suffix}")
+
+
+def get_original_path(path: PathType) -> Path:
+    _path = Path(path)
+    filename = _path.stem
+    if filename.startswith("."):
+        filename = f"{Path(Path(filename[1:]).stem).stem}{_path.suffix}"
```
mdlu, line 29: this is not a legible line of code 🤣 but looks ok

src/allotropy/allotrope/schema_parser/generate_schemas.py
```diff
@@ -13,30 +16,29 @@
...
+from allotropy.allotrope.schemas import get_schema
 
 
 def lint_file(model_path: Path) -> None:
     # The first run of ruff changes typing annotations and causes unused imports. We catch failure
     # due to unused imports.
     try:
         subprocess.check_call(
-            f"ruff {model_path} --fix",
+            f"ruff --ignore N999 {model_path} --fix",
```
mdlu, line 41: why do we need this?

src/allotropy/allotrope/schema_parser/model_class_editor.py
```diff
@@ -45,9 +48,11 @@ def _schemas_equal(schema1: dict[str, Any], schema2: dict[str, Any]) -> bool:
...
 
 
-def get_shared_schema_info(schema_path: Path) -> tuple[set[str], dict[str, set[str]]]:
-    with open(schema_path) as f:
-        schema = json.load(f)
+def get_shared_schema_info(
+    schema: dict[str, Any]
+) -> tuple[set[str], dict[str, set[str]]]:
+    # with open(schema_path) as f:
+    #    schema = json.load(f)
```
mdlu, line 55: delete comments?

src/allotropy/allotrope/schema_parser/path_util.py
```diff
@@ -67,9 +81,10 @@ def get_schema_path_from_reference(reference: str) -> Path:
     return Path(f"{ref_match.groups()[0]}.json")
 
 
-def get_model_file_from_schema_path(schema_path: Path) -> Path:
+def get_model_path_from_schema_path(schema_path: Path) -> Path:
     rel_schema_path = PureWindowsPath(get_rel_schema_path(schema_path))
     schema_file = rel_schema_path.name
+    # TODO: remove lower?
```
mdlu, line 87: why/why not?

src/allotropy/allotrope/schema_parser/path_util.py
```diff
@@ -89,7 +117,7 @@ def get_import_path_from_path(model_path: Path) -> str:
 
 def get_model_class_from_schema(asm: Mapping[str, Any]) -> Any:
     schema_path = get_schema_path_from_manifest(asm["$asm.manifest"])
-    model_file = get_model_file_from_schema_path(Path(schema_path))
+    model_file = get_model_path_from_schema_path(Path(schema_path))
```
mdlu, line 120: nit: should be `model_path` now, right

src/allotropy/allotrope/schema_parser/update_units.py
```diff
@@ -47,7 +44,7 @@ class {unit_name}(HasUnit):
     unit: str = {UNIT_STRING_OVERRIDES.get(unit, unit_str)}"""
             )
 
-    with open(CUSTOM_MODELS_PATH, "w") as f:
```
mdlu, line 50: just checking, are `CUSTOM_MODELS_PATH` and `UNITS_SCHEMAS_PATH` still used now, with this change? (should we remove?)

src/allotropy/parsers/utils/pandas.py
```diff
@@ -80,6 +81,14 @@ def assert_value_from_df(df: pd.DataFrame, key: str) -> Any:
         raise AllotropeConversionError(msg) from e
 
 
+def _normalize_columns(df: pd.DataFrame) -> pd.DataFrame:
```
mdlu, line 84: I feel like I just reviewed these changes, right? (will probably be fixed once the other PR merges in first, I assume just some commit got mixed in)

mdlu (dismissed):

src/allotropy/allotrope/schema_parser/backup_manager.py
nathan-stender: Who me? Write illegible code?!? I never :P

nathan-stender: (added a comment)

src/allotropy/allotrope/schema_parser/generate_schemas.py
nathan-stender: Added a comment. N999 checks that the filename is valid for a python module, which fails here because we're now calling with the backup file (which has .uuid.bak in the path)

src/allotropy/allotrope/schema_parser/path_util.py
nathan-stender: Oh oops, I leave these to follow up myself before putting up PR 👀 . For the record - it was removed because it wasn't needed for the model filename, which is never capitalized for either the model or the schema

It is needed below this line for the `REC` -> `rec` transform when converting a schema path to a model path

src/allotropy/allotrope/schema_parser/update_units.py
nathan-stender: Hmmm good point, they are technically still used in `GENERATED_SHARED_PATHS`, which is backed up and then passed to the unit update code, but no reason to have separate constants. Fixed.

src/allotropy/parsers/utils/pandas.py
nathan-stender: Oh dang yep, must not have switched back to main before opening this up. That merged, will update.

stephenworlow (approved):

PR #597, opened 2024-08-18
refactor: Use templates to simplify parser syntax

nathan-stender:
This PR:

1. Adds a base class for SchemaMappers, factoring out constructor
2. Adds MapperVendorParser, which allows parsers using a schema mapper to specify a `SCHEMA_MAPPER` and implement `create_data` to automatically map data to model
3. Implements a `create_data` function for each `MapperVendorParser`, often pulling some of the implementation out of the structure class
4. Changes `parser_name` and `release_state` property to constants (for simplicity)

These changes further reduce the boilerplate code for parsers, and establish a consistent pattern by enforcing the implementation of create_data.

A new pattern introduced here is that the parser creates the `Data` class for the schema mapper, where previously this was mostly done in a `create_data` function in the `structure` class. The thinking here is that the parser is responsible for being the "bridge" between the reader, structure, and mapper classes.

Finally, this PR does some other misc cleanup refactors to continue reinforcing patterns, e.g.:

- adding a couple reader classes where raw data file data was processed in a structure/parser class
- adding constants files where they were missing (and constants were embedded)


rayaankha-tara (approved):
nice!

PR #609, opened 2024-08-24
test: Update all real uuids with test uuids in tests

nathan-stender:


tests/parsers/agilent_tapestation_analysis/agilent_tapestation_analysis_structure_test.py
```diff
@@ -185,73 +182,71 @@ def testcreate_measurement_groups() -> None:
...
+    with mock_uuid_generation():
         _, calc_docs = create_measurement_groups(
             get_samples_xml(with_calculated_data=True)
         )
 
-    peak_data_source = DataSource(feature="peak", identifier="dummy_id")
-    sample_data_source = DataSource(feature="sample", identifier="dummy_id")
+    sample_data_source = DataSource(feature="sample", identifier="TEST_ID_0")
+    peak_1_data_source = DataSource(feature="peak", identifier="TEST_ID_2")
+    peak_2_data_source = DataSource(feature="peak", identifier="TEST_ID_7")
```
stephenworlow, line 192: how did you determine the number ordering of the variables?!

tests/parsers/agilent_tapestation_analysis/agilent_tapestation_analysis_structure_test.py
```diff
@@ -288,7 +281,7 @@ def testcreate_measurement_groups_with_regions() -> None:
             comment="Partially Degraded",
         ),
         ProcessedDataFeature(
-            identifier="dummy_id",
+            identifier="TEST_ID_6",
```
stephenworlow, line 284: is there a rhyme or reason behind the TEST_ID_X? It does seem random

stephenworlow (approved):

tests/parsers/agilent_tapestation_analysis/agilent_tapestation_analysis_structure_test.py
nathan-stender: The incrementing ID is really just a way to make the fake uuids distinct. We could instead use a consistent seed for uuid generation, but I think it would be harder to trace around.

For what it's worth, if you're familiar with a particular parser, the IDs usually make sense to you.

In this, for example, the parser generates all the processed data and calculated data items for a particular measurement, then moves to the next one.

That's why we get processed id 3, then calculated ids 4 & 5, then processed id 6.

stephenworlow (approved):

PR #617, opened 2024-08-26
feat: Add supported extensions to Vendor/parser

nathan-stender:
This allows us to always throw a helpful error if an invalid file type is passed, and will allow users to tell which filetypes are supported.


stephenworlow (approved):
yayyyyy

PR #619, opened 2024-08-27
refactor: Move detection_type to Measurement

nathan-stender:
Because `detection_type` can be different for different measurements in a document, move detection_type to Measurement

### DIFF
--- a/src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
+++ b/src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
@@ -52,7 +52,6 @@ from allotropy.allotrope.models.shared.definitions.definitions import (
 from allotropy.allotrope.schema_mappers.schema_mapper import SchemaMapper
 from allotropy.constants import ASM_CONVERTER_VERSION
 from allotropy.exceptions import AllotropyParserError
-from allotropy.parsers.agilent_gen5.constants import ReadMode
 from allotropy.parsers.utils.values import assert_not_none, quantity_or_none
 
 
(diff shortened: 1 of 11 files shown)


src/allotropy/parsers/example_weyland_yutani/constants.py
```diff
@@ -1,4 +1,5 @@
 DEVICE_TYPE = "fluorescence detector"
+DETECTION_TYPE = "Fluorescence"
```
mdlu, line 2: weyland yutani strikes again 🤣

src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
```diff
@@ -394,7 +393,7 @@ def _get_fluorescence_measurement_document(
                 device_control_document=[
                     FluorescencePointDetectionDeviceControlDocumentItem(
                         device_type=metadata.device_type,
-                        detection_type=ReadMode.FLUORESCENCE.value,
+                        detection_type=measurement.detection_type,
```
mdlu, line 396: mostly a curiosity question, it is surprising to me that a fluorescence measurement document isn't always of type `FLUORESCENCE` (and likewise for the other modes) for example, if a `FluorescencePointDetectionDeviceControlDocumentItem` is actually of type `LUMINESCENCE` or something, isn't it going to be missing some fields that it needs that are in `LuminescencePointDetectionDeviceControlDocumentItem`?

mdlu (approved):

src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
nathan-stender: I had the same thought - I originally thought that in this PR I was going to get rid of `MeasurementType` and use `detection type` to determine the type of measurement document to create.

Looking into the existing implementation though, sometimes detection_type is more specific (which is when it really makes sense to even have this field). For example, in the `Revvity Kaleido` parser, we read the detection type from the input data experiment type.

PR #621, opened 2024-08-27
fix: Mabtech Apex, Update regex to handle scenarios where first word in the machine ID section is all uppercase

joshua-benchling:
When testing an Mabtech Apex example file we are getting this error "Unable to interpret Machine ID" because our parser is expecting 'Iris' as value but the export value is 'IRIS'


src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
```diff
@@ -21,7 +21,7 @@
 def create_metadata(plate_info: SeriesData, file_name: str) -> Metadata:
     machine_id = assert_not_none(
         re.match(
-            "([A-Z]+[a-z]+) ([0-9]+)",
+            "([A-Z]+[a-z]*|[A-Z]+) ([0-9]+)",
```
stephenworlow, line 24: Shouldn't just changing to "([A-Z]+[a-z]*) ([0-9]+)" be sufficient?

rayaankha-tara (approved):

stephenworlow (approved):

PR #626, opened 2024-08-29
refactor: Use schema mapper for revvity kaleido parser

nathan-stender:


src/allotropy/parsers/agilent_gen5_image/agilent_gen5_image_structure.py
```diff
@@ -350,6 +351,7 @@ def _create_measurement(
         detector_gain_setting=instrument_settings.detector_gain,
         magnification_setting=read_section.magnification_setting,
         illumination_setting=instrument_settings.illumination,
+        illumination_setting_unit=UNITLESS,
```
nathan-stender, line 354: NOTE: illumination can be unitless or %, revvity uses %, so we specify unit now.

mdlu (dismissed):

stephenworlow (approved):

PR #630, opened 2024-08-30
feat: update quantity calculated documents of AppBio Quantstudio

alejandro-salgado:
- update quantity calculated documents of AppBio Quantstudio by adding Y-intercept and Slope as data sources


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_calculated_documents.py
```diff
@@ -30,6 +30,8 @@ def build_quantity(well_item: WellItem) -> CalculatedDocument | None:
         value=quantity,
         data_sources=[
             DataSource(feature="cycle threshold result", reference=well_item),
+            DataSource(feature="Y-intercept", reference=well_item),
+            DataSource(feature="Slope", reference=well_item),
```
mdlu, line 34: question: seems like the feature names aren't capitalized elsewhere, should it be `y-intercept` and `slope` throughout this PR, to match?

alejandro-salgado: yes, you are right. I change them

tests/parsers/appbio_quantstudio/testdata/appbio_quantstudio_multiple_cal_doc_wells.json
```diff
@@ -495,6 +496,14 @@
                             {
                                 "data source identifier": "APPBIO_QUANTSTUDIO_TEST_ID_0",
                                 "data source feature": "cycle threshold result"
+                            },
```
nathan-stender, line 499: How are we "referencing" y-intercept and slope when they do not appear in the measurements? 

I see in the test files that there are columns for `y-intercept` and `slope` - shouldn't we add them to the processed data for this to make sense?

nathan-stender (changes requested):

tests/parsers/appbio_quantstudio/testdata/appbio_quantstudio_multiple_cal_doc_wells.json
alejandro-salgado: Added check to add data sources only when their values are non None.

mdlu (approved):

tests/parsers/appbio_quantstudio/testdata/appbio_quantstudio_multiple_cal_doc_wells.json
nathan-stender: @alejandro-salgado I'm still not clear that this is addressed. My point is not that the measurement could be None, my point is that the `y-intercept` and `slope` do not appear in the data source. There is no entry for `y-intercept` and `slope` in the measurement's processed data. I think you need to add them to the processed data?

PR #631, opened 2024-08-30
feat: Add Unchained labs lunatic parser support for all formats with all file types

nathan-stender:


tests/parsers/unchained_labs_lunatic/testdata/Example_Lunatic_Plate_Reader_csv_no_header.csv
```diff
@@ -0,0 +1,5 @@
+Plate ID,"Plate
+Position",Sample name,"Concentration
+(ng/ul)",A260/A280,A260/A230,A260,Sample group,Date,Time,Application,Instrument ID
```
mdlu, line 3: the fact that files can show up with a header like this makes me upset 😭 but glad we handle it

mdlu (approved):

PR #633, opened 2024-08-30
feat: Agilent Gen5 Kinetic Measurements Support

sumaga-benchling:


nathan-stender (changes requested):
Some initial comments, requesting a bit of restructuring for integration with schema mappers. 

I will do a more thorough review of parser/structure changes after these changes are addressed.

src/allotropy/allotrope/schema_mappers/adm/plate_reader/rec/_2024/_06/plate_reader.py
```diff
@@ -242,15 +264,24 @@ def _get_ultraviolet_absorbance_measurement_document(
...
                     )
                 ]
             ),
-            absorbance=TQuantityValueMilliAbsorbanceUnit(
-                value=assert_not_none(
-                    value=measurement.absorbance,
-                    msg="Missing absorbance value in ultraviolet absorbance measurement",
-                )
-            ),
+            absorbance=TQuantityValueMilliAbsorbanceUnit(value=measurement.absorbance)
```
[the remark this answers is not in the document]
joshua-benchling: Thanks for pointing this out, I completely forgot about point-detector schema

joshua-benchling (approved):

src/allotropy/allotrope/schema_mappers/adm/plate_reader/rec/_2024/_06/plate_reader.py
```diff
@@ -403,3 +510,28 @@ def _get_calculated_data_aggregate_document(
...
+                    TDatacubeComponent(
+                        field_componentDatatype=profile_data_cube.structure_measures.type_,
+                        concept=profile_data_cube.structure_measures.concept,
+                        unit=profile_data_cube.structure_measures.unit,
+                    )
+                ],
+            ),
+            data=TDatacubeData(
+                dimensions=[profile_data_cube.dimensions],
+                measures=[profile_data_cube.measures],
```
slopez-b, line 535: these two are already lists.
```suggestion
                dimensions=profile_data_cube.dimensions,
                measures=profile_data_cube.measures,
```

src/allotropy/allotrope/schema_mappers/adm/plate_reader/rec/_2024/_06/plate_reader.py
```diff
@@ -67,6 +78,25 @@ class MeasurementType(Enum):
...
+    type_: FieldComponentDatatype
+    concept: str
+    unit: str
+
+
+@dataclass
+class DataCube:
+    label: str
+    structure_dimensions: DataCubeComponent
+    structure_measures: DataCubeComponent
```
slopez-b, line 97: These should be lists, since we can get multiple structure dimentions and measures per datacube.

tests/parsers/agilent_gen5/testdata/absorbance/240411_172731_BNCH2345883_abs450_96well_non_numeric_values.json
```diff
@@ -40,21 +35,20 @@
                             }
                         }
                     ],
+                    "measurement time": "2024-04-11T17:27:15+00:00",
+                    "plate well count": {
+                        "value": 83,
```
nathan-stender, line 40: This is correct, even when a value doesn't have a valid measurement, plate_well_count should not change.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -238,9 +271,8 @@ def create(cls, lines: list[str]) -> list[ReadData]:
         read_type = cls.get_read_type(procedure_details)
         if read_type != ReadType.ENDPOINT:
             raise AllotropeConversionError(UNSUPPORTED_READ_TYPE_ERROR)
-
         read_modes = cls.get_read_modes(procedure_details)
-        read_sections = list(SectionLinesReader(lines).iter_sections("^Read\t"))
+        read_sections = list(SectionLinesReader(lines).iter_sections(r"^\s{0,4}Read\t"))
```
nathan-stender, line 275: I'm guessing `^\s{0,4}` is to match a variable number of whitespace. Let's do `^\s*` to cover all cases.

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -453,10 +483,29 @@ def get_temperature(actual_temperature_lines: list[str] | None) -> float | None:
...
+    data = read_csv(StringIO("\n".join(kinetic_lines)), sep="\t")
+    kinetic_measurements: defaultdict[str, list[float | None]] = defaultdict(
+        list[float | None]
+    )
+    kinetic_elapsed_time: list[float] = []
+    for col in data.columns:
+        for _idx, value in enumerate(data[col]):
+            if col == "Time":
+                kinetic_elapsed_time.append(_convert_time_to_seconds(str(value)))
+            elif not pd.isna(value):
```
nathan-stender, line 500: This isn't the correct way to do this - you are constructing lists from the columns of a dataframe. If you drop NA values in the middle of a column, you will end up with mis-aligned data.

Looking at the sample data, I think you are doing this because there is an entire column with NA values. So what we really want to do is drop columns that are all NA, and then error is a column has some NA and some not. 

Note that you can drop all null columns by calling `.dropna(axis="columns", how="all")` when creating the data

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -453,10 +483,29 @@ def get_temperature(actual_temperature_lines: list[str] | None) -> float | None:
...
+) -> tuple[dict[str, list[float | None]], list[float]] | None:
+    if not kinetic_lines:
+        return None
+    data = read_csv(StringIO("\n".join(kinetic_lines)), sep="\t")
+    kinetic_measurements: defaultdict[str, list[float | None]] = defaultdict(
+        list[float | None]
+    )
+    kinetic_elapsed_time: list[float] = []
+    for col in data.columns:
+        for _idx, value in enumerate(data[col]):
```
nathan-stender, line 497: With regard to this iteration, I think you can simplify this a bit using `pandas` `to_list()`:

```
    kinetic_elapsed_time = data["Time"].map(lambda val: _convert_time_to_seconds(str(val))).to_list()
    for col_name, column in data.items():
        if col_name == "Time":
            continue
        if column.isnull().any():
            msg = f"Unable to process null value in column: {col_name} of Kinetic section."
        kinetic_measurements[str(col_name)] = column.astype(float).to_list()
    return dict(kinetic_measurements.items()), kinetic_elapsed_time
```

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -585,31 +723,57 @@ def create_metadata(header_data: HeaderData) -> Metadata:
...
-    if read_data.read_mode == ReadMode.ABSORBANCE:
-        measurement_type = MeasurementType.ULTRAVIOLET_ABSORBANCE
-    elif read_data.read_mode == ReadMode.FLUORESCENCE:
-        measurement_type = MeasurementType.FLUORESCENCE
-    elif read_data.read_mode == ReadMode.LUMINESCENCE:
-        measurement_type = MeasurementType.LUMINESCENCE
-
-    detector_wavelength_setting: JsonFloat | None = None
-    if measurement_type is MeasurementType.ULTRAVIOLET_ABSORBANCE:
+    match (read_data.read_mode, bool(kinetic_data)):
```
nathan-stender, line 755: Very sorry you went through changing this to a `match` statement, but we need to keep it `if` for now.

The problem is that Benchling can't use this syntax yet, not that it won't work here.

nathan-stender (changes requested):

nathan-stender:
Adding @james-leinas review for the new data cube test case generated

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
joshua-benchling: ohhh I see, np I can revert.

tests/parsers/agilent_gen5/testdata/absorbance/240411_172731_BNCH2345883_abs450_96well_non_numeric_values.json
james-leinas: @joshua-benchling agree with Nate's comment here. We should obtain count of all the wells even if no actual measurement from some.

tests/parsers/agilent_gen5/testdata/absorbance/010307_114129_BNCH654563_stdcurve_singleplate01.json
```diff
@@ -40,21 +35,20 @@
                             }
                         }
                     ],
+                    "measurement time": "2022-10-10T21:00:04+00:00",
+                    "plate well count": {
+                        "value": 288,
```
nathan-stender, line 40: Plate well count is still not correct - this value should not change from the previous implementation. 

Looks like you need to count the number of well positions, not the number of measurements?

nathan-stender (changes requested):

tests/parsers/agilent_gen5/testdata/absorbance/010307_114129_BNCH654563_stdcurve_singleplate01.json
james-leinas: @joshua-benchling, agree with Nate here. We should count number of positions rather than measurements. Likely expected values for the field should be 96 or 384 in most cases.

nathan-stender (approved):
🎉

james-leinas (approved):

PR #634, opened 2024-09-03
fix: Update Mabtech apex adapter to support LED filter

joshua-benchling:
Update Mabtech apex adapter to support LED filter and add new custom information document fields


src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
```diff
@@ -50,19 +60,34 @@ def _create_measurement(plate_data: SeriesData) -> Measurement:
...
                 for feature in constants.IMAGE_FEATURES
             ],
         ),
+        custom_information_document=CustomInformationDocument(
+            led_filter=plate_data.get(str, "LED Filter") or NOT_APPLICABLE
+        ),
+    )
+
+
+def _build_feature(
```
nathan-stender, line 78: Can you instead make this `_format_feature_name` and just return the feature string (leaving the `ImageFeature` logic above) this makes the function scope as small and clear as possible.

src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
```diff
@@ -50,19 +60,34 @@ def _create_measurement(plate_data: SeriesData) -> Measurement:
...
-                    identifier=random_uuid_str(),
-                    feature=feature,
-                    result=plate_data.get(float, feature, NaN),
-                )
+                _build_feature(feature, plate_data, led_filter)
                 for feature in constants.IMAGE_FEATURES
             ],
         ),
+        custom_information_document=CustomInformationDocument(
+            led_filter=plate_data.get(str, "LED Filter") or NOT_APPLICABLE
```
nathan-stender, line 73: You already retrieved `LED Filter` above, so use it here. 

Also - it seems like we should leave this out if it is not provided, instead of "N/A". If this does not agree with a requirement you got, let me know.

src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
```diff
@@ -50,19 +60,34 @@ def _create_measurement(plate_data: SeriesData) -> Measurement:
...
+    if led_filter:
+        led_number = re.search(r"\d+", led_filter)
+        if led_number:
+            feature = feature.format(filter=led_filter, led_number=led_number.group())
+    else:
+        feature = feature.format(filter="", led_number="").replace("  ", " ")
+    return ImageFeature(
+        identifier=random_uuid_str(),
+        feature=feature,
+        result=plate_data.get(float, feature, NaN),
```
nathan-stender, line 90: I know this is existing logic, but it seems like we have added more values that are missing for many of the test cases. It seems like we should instead skip creating an ImageFeature if the value is not found?

nathan-stender (changes requested):

src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
joshua-benchling: asking James about it..

src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
```diff
@@ -50,19 +60,34 @@ def _create_measurement(plate_data: SeriesData) -> Measurement:
...
+        ),
+    )
+
+
+def _build_feature(
+    feature: str, plate_data: SeriesData, led_filter: str | None
+) -> ImageFeature:
+    if led_filter:
+        led_number = re.search(r"\d+", led_filter)
+        if led_number:
```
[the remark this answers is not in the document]
joshua-benchling: yes, makes sense to me.

src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
```diff
@@ -144,6 +147,14 @@ class Measurement:
     image_count_setting: float | None = None
     fluorescent_tag_setting: str | None = None
 
+    # Custom information
+    custom_information_document: CustomInformationDocument | None = None
+
+
+@dataclass(frozen=True)
+class CustomInformationDocument:
```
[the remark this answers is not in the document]
joshua-benchling: Sounds good to me, this was actually my initial approach.

nathan-stender (changes requested):

src/allotropy/allotrope/schema_mappers/adm/plate_reader/benchling/_2023/_09/plate_reader.py
```diff
@@ -440,11 +456,15 @@ def _get_sample_document(self, measurement: Measurement) -> SampleDocument:
...
 
     def _get_processed_data_aggregate_document(
-        self, data: ProcessedData | None
+        self, data: ProcessedData | None, well_plate_identifier: str | None = None
     ) -> ProcessedDataAggregateDocument | None:
         if not data:
             return None
 
+        if not data.features:
+            logging.warning(f"no image features identified for {well_plate_identifier}")
```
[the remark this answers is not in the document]
joshua-benchling: Hey Nate, thanks for pointing this out.
Yes, this behavior is specific to Mabtech apex adapter so I am moving the warning there.
We are not skipping the measurement when there are not features we are just skipping the processed data document.

src/allotropy/parsers/mabtech_apex/mabtech_apex_structure.py
```diff
@@ -41,29 +43,65 @@ def create_metadata(plate_info: SeriesData, file_name: str) -> Metadata:
...
+) -> ImageFeature | None:
+    if led_filter:
+        led_number = re.search(r"\d+", led_filter)
+        if not led_number:
+            error_msg = f"Unable to interpret LED number from {led_filter}"
+            raise AllotropeParsingError(error_msg)
+        feature = feature.format(filter=led_filter, led_number=led_number.group())
+    else:
+        feature = feature.format(filter="", led_number="").replace("  ", " ")
+    if not plate_data.get(float, feature):
```
nathan-stender, line 98: Need to check if value `is not None`, see test data where you drop some values that are 0.

Also, you can fetch the value here and use it below:

```
if (value := plate_data.get(float, feature)) is None:
    return None
```

nathan-stender (changes requested):

ajcariaga16 (approved):

nathan-stender (changes requested):
I'm not sure exactly what is going on, but you are creating multiple duplicate measurements in this test data. I think that's why you had to adjust to have multiple measurements per row

nathan-stender (approved):
Discussed offline - I see that the "repeat measurements" have different labels, and do in fact have identical data in the source file, so this is working correctly.

PR #635, opened 2024-09-03
feat: additional softmaxpro changes

slopez-b:
Additional changes to the `Softmax Pro` Adapter:

- Added error message for zero plate reader documents in `Softmax Pro` Adapter
- Disregard compartment temperature in `Softmax Pro` adapter when is reported as 0.

Note: The compartment temperature was included in the `Softmax Pro`  output even when it was reported as `0` because this was a required field in a previous version of the schema. Since `REC/2024/06 plate reader` it is no longer required.


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_parser.py
```diff
@@ -149,7 +151,9 @@ def _get_fluorescence_measurement_document(
                 ),
                 compartment_temperature=quantity_or_none(
                     TQuantityValueDegreeCelsius,
-                    data_element.temperature,
+                    # A temperature of 0 degrees indicates an error in the instrument Software
```
nathan-stender, line 154: Add you add a TODO to add an error document?

nathan-stender (approved):

PR #662, opened 2024-09-20
test: Remove unit tests that just call parser end-to-end

nathan-stender:
Remove tests that are just re-creating `to_allotropy_test` but manually specifying the data classes.

These do not add coverage, and are hard to maintain.

Note that this does not delete tests that exercise specific functions of a parser/structure class.


stephenworlow (dismissed):

stephenworlow (dismissed):

src/allotropy/testing/utils.py
```diff
@@ -196,3 +190,9 @@ def validate_contents(
                 msg = f"Mismatch between actual and expected for '{expected_file}', writing expected output because 'write_actual_to_expected_on_fail=True'\n\n{e}"
                 raise AssertionError(msg) from e
         raise
+
+    # Ensure that allotrope_dict can be structured back into a python model.
```
nathan-stender, line 194: Note: this is so if a test fails with different content than on local, we see the diff in content before other failures.

stephenworlow (approved):

PR #665, opened 2024-09-24
feat: Chemometec NC View Adapter

joshua-benchling:
- Added Chemometec NC View Adapter


src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_reader.py
```diff
@@ -0,0 +1,18 @@
...
+)
+
+
+class ChemometecNcViewReader:
+    SUPPORTED_EXTENSIONS = "xlsx,csv"
+    header: SeriesData
+    data: pd.DataFrame
+
+    def __init__(self, named_file_contents: NamedFileContents) -> None:
+        self.data = read_csv(named_file_contents.contents)
```
nathan-stender, line 17: What if the data is xlsx? I think I remember this discussion from standup - if we don't have an xlsx example, you can just remove it from supported extensions for now.

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_structure.py
```diff
@@ -0,0 +1,54 @@
...
+    return MeasurementGroup(
+        analyst=data[str, "OPERATOR"],
+        measurements=[
+            Measurement(
+                measurement_identifier=random_uuid_str(),
+                processed_data_identifier=random_uuid_str(),
+                timestamp=_format_timestamp(data[str, "NAME"]),
+                sample_identifier=data[str, "SAMPLE ID"],
+                viability=data[float, "VIABILITY (%)"],
+                total_cell_density=_format_unit(data[str, "TOTAL (cells/ml)"]),
```
nathan-stender, line 37: Total cell density is not required in the ASM document, for that reason, we should make this `data.get`

I believe this probably applies for some other fields in this block.

Generally, if the `Measurement` field is not required, the retrieval should not be required.

nathan-stender (changes requested):
Excellent work! Code LGTM - two things to address

james-leinas (approved):

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_parser.py
```diff
@@ -0,0 +1,33 @@
...
+)
+from allotropy.parsers.chemometec_nc_view.constants import DISPLAY_NAME
+from allotropy.parsers.release_state import ReleaseState
+from allotropy.parsers.utils.pandas import map_rows
+from allotropy.parsers.vendor_parser import MapperVendorParser
+
+
+class ChemometecNcViewParser(MapperVendorParser[Data, Model]):
+    DISPLAY_NAME = DISPLAY_NAME
+    RELEASE_STATE = ReleaseState.WORKING_DRAFT
```
nathan-stender, line 24: I thought this was to be updated to RECOMMENDED?

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_structure.py
```diff
@@ -0,0 +1,60 @@
...
+                    data.get(str, "AGGREGATES (%)")
+                ),
+                debris_index=data.get(float, "DEBRIS INDEX"),
+                cell_density_dilution_factor=data.get(float, "DILUTION FACTOR"),
+            )
+        ],
+    )
+
+
+def _format_number(unit: str | None) -> JsonFloat:
```
nathan-stender, line 53: Hmm, I think we should actually return None here if the value is None, or the value fails to parse. I didn't see anything about NaN in the req doc?

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_structure.py
```diff
@@ -0,0 +1,60 @@
...
+    return MeasurementGroup(
+        analyst=data[str, "OPERATOR"],
+        measurements=[
+            Measurement(
+                measurement_identifier=random_uuid_str(),
+                processed_data_identifier=random_uuid_str(),
+                timestamp=_format_timestamp(data[str, "NAME"]),
+                sample_identifier=data[str, "SAMPLE ID"],
+                viability=data[float, "VIABILITY (%)"],
+                total_cell_density=_format_number(data.get(str, "TOTAL (cells/ml)")),
```
nathan-stender, line 37: Just realized - these values need to be converted to `10^6 cells/mL` (millions of cells / mL) that's what we report in ASM.

To do this you can just divide by 10^6 if not None.

(see cedex hires for an example)

nathan-stender (changes requested):

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_parser.py
joshua-benchling: For some reason a unitest was failing when I had it in REC state, I just try again and it worked!

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_structure.py
joshua-benchling: right, this should apply for optional arguments, I made another method to cover the case for `viable_cell_density` which should be required and could be NaN.

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_structure.py
```diff
@@ -0,0 +1,77 @@
...
+            )
+        ],
+    )
+
+
+def _format_number(unit: str | None) -> JsonFloat | None:
+    if not unit:
+        return None
+    number = try_float_or_none("".join(unit.split()))
+    return float(Decimal(number) / Decimal("1000000")) if number else None
```
nathan-stender, line 64: Sorry, but can you retrieve the values without dividing and then do the division for the fields?

I am worried about adding the division to this generic function, because if someone used it for something other than a cell density number it would incorrectly divide.

src/allotropy/parsers/chemometec_nc_view/chemometec_nc_view_structure.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _format_required_number(unit: str | None) -> JsonFloat:
+    if not unit:
+        return NaN
```
nathan-stender, line 69: Also just want to check here - are we sure we should be returning NaN instead of erroring for a bad required number?

nathan-stender (changes requested):

nathan-stender (approved):

james-leinas (approved):

nathan-stender (approved):

PR #679, opened 2024-09-27
feat: Agilent Gen5 Image adapter, add support for no result sections file examples

joshua-benchling:
Added support for agilent gen5 image files where there is not a results section


src/allotropy/parsers/agilent_gen5_image/agilent_gen5_image_structure.py
```diff
@@ -250,11 +250,32 @@ def _get_read_type(cls, procedure_details: str) -> ReadType:
...
+    result_lines: list[str] | None,
     header_data: HeaderData,
     read_data: ReadData,
     sample_identifiers: dict[str, str],
 ) -> list[MeasurementGroup]:
+    if not result_lines:
+        return [
+            MeasurementGroup(
+                measurement_time=header_data.datetime,
+                plate_well_count=0,
```
nathan-stender, line 262: I think we can still read plate well count. I see 96 in "Plate Type" in the header

Perhaps we need to adjust our plate_well_count strategy for this adapter to read from there.

src/allotropy/parsers/agilent_gen5_image/agilent_gen5_image_structure.py
```diff
@@ -250,11 +250,32 @@ def _get_read_type(cls, procedure_details: str) -> ReadType:
...
+                analytical_method_identifier=header_data.protocol_file_path,
+                experimental_data_identifier=header_data.experiment_file_path,
+                measurements=[
+                    _create_measurement(
+                        "N/A",
```
nathan-stender, line 267:
```suggestion
                        NOT_APPLICABLE,
```
(import constant)

src/allotropy/parsers/agilent_gen5_image/agilent_gen5_image_structure.py
```diff
@@ -250,11 +250,32 @@ def _get_read_type(cls, procedure_details: str) -> ReadType:
...
+                        header_data,
+                        read_section,
+                        instrument_settings,
+                        None,
```
nathan-stender, line 271: Let's set `sample identifier` to `METATADA ONLY` (@james-leinas to review wording)

nathan-stender (changes requested):
Mostly LGTM - couple requests

src/allotropy/parsers/agilent_gen5_image/agilent_gen5_image_structure.py
james-leinas: @nathan-stender, believe that we decided not to use "Plate Type" as the plate well count reference since the instrument file can have other vendor specific names for plates that don't include the number of wells. 

Logic could be for these examples is to look for well number if present in "Plate Type" field, record "N/A" if not found. Think best approach is still to pull from the plate well matrix, when available.

james-leinas: I see issue with "N/A" now being that the field requires numeric value. Have to think about that implementation

nathan-stender: @james-leinas I thought what we landed on was a fallback strategy:

1. Read from plate data matrix (current approach) if at all possible
2. Try to read from Plate Type if there is a number string in the plate type. Given that this is customer set and we don't want to record any-old thing (like if they name the plate `Plate type 2` or something we don't want to pull 2) perhaps this can just search for some expected numbers, like 96 and 384?
3. Record 0, as (@joshua-benchling this is a numeric-only field in ASM) - honestly at this point I doubt it's that big of a deal, for this use case customer probably doesn't care about it?

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
```diff
@@ -87,8 +89,16 @@ def create(cls, data: SeriesData, file_name: str) -> HeaderData:
             well_plate_identifier=plate_identifier or data.get(str, "Plate Number"),
             model_number=data.get(str, "Reader Type:"),
             equipment_serial_number=data.get(str, "Reader Serial Number:"),
+            plate_well_count=plate_well_count,
         )
 
+    @staticmethod
+    def _extract_well_count(plate_type: str) -> float:
+        match = re.search(r"\d+", plate_type)
```
nathan-stender, line 97: Let's search for a specific list of possible well counts, to avoid using a random number identifier

```suggestion
        match = re.search(rf"({'|'.join(str(count) for count in constants.POSSIBLE_WELL_COUNTS)})", plate_type).groups()
```

Where `constants.POSSIBLE_WELL_COUNTS = [48, 96, 384, 1536]`

cc: @james-leinas to approve

nathan-stender (changes requested):
Code LGTM - one last safety check

src/allotropy/parsers/agilent_gen5/agilent_gen5_structure.py
james-leinas: @nathan-stender, here would be my suggested list of possible well counts [1, 2, 4, 6, 8, 12, 24, 48, 72, 96, 384, 1536, 3456]. Gen5/Gen5 Image offers a lot of compatibility of plate types. Agree that this will avoid arbitrary numbers being flagged for plate well count.

joshua-benchling: Sounds good to me

james-leinas (approved):

nathan-stender (approved):

PR #680, opened 2024-09-27
feat: Unchained labs - get additional metadata from header block

slopez-b:
Grab additional metadata from header block if present

Software version - > software version  (already being mapped, added to tests)
Test performed by -> analyst
Client version -> firmware version
Experiment name -> experimental data identifier


nathan-stender (approved):
Code LGTM!

Can you change the title to:

`feat: Unchained Labs Lunatic - add additional metadata from header block in xlsx format`

PR #688, opened 2024-09-30
feat: NanoDrop 8000 & NanoDrop Eight - support alternative reporting format for absorbance measurements

nathan-stender:
In the alternative format, instead of reporting absorbances in columns names for the wavelength (`A260`) they are recorded as `abs <id>`/`nm <id>` pairs.

Also updated comments around how we store mass concentration to clarify a bit.


nathan-stender:
For anyone watching - putting this back in draft while I work on one more fix for this adapter

slopez-b (approved):
LGTM

PR #689, opened 2024-10-01
fix: AppBio Quantstudio - omit wells that have no results in the raw data file, indicating an omitted well

nathan-stender:


src/allotropy/parsers/appbio_quantstudio_designandanalysis/appbio_quantstudio_designandanalysis_calculated_documents.py
```diff
@@ -738,7 +738,7 @@ def iter_relative_standard_curve_calc_docs(
             build_relative_rq_min(view_st_data, view_tr_data, sample, target)
         )
         calc_docs.append(
-            build_relative_rq_min(view_st_data, view_tr_data, sample, target)
+            build_relative_rq_max(view_st_data, view_tr_data, sample, target)
```
mdlu, line 741: oops was this just a bug haha (wasn't covered by an existing test?)

mdlu (approved):
just to confirm that the desired behavior is to entirely skip wells without results, rather than adding some sort of "placeholder value" indicating it was without results? (getting flashbacks to my recent empty wells lookup config ticket haha)

nathan-stender:
@mdlu yeah that's the approach we're taking right now. Might figure out a way to support empty well metadata in the future, but for now we just want to unblock customers with this use case.

PR #690, opened 2024-10-01
feat: Update Methodical Mind to use the plate reader REC schema

joshua-benchling:
* Changed schema mapper
* Added new fields in the data system document.


nathan-stender (approved):

src/allotropy/parsers/methodical_mind/methodical_mind_structure.py
```diff
@@ -87,12 +87,14 @@ def create(luminescence: int, location_id: str, well_plate_id: str) -> WellData:
 def create_metadata(header: Header) -> Metadata:
     return Metadata(
         file_name=header.file_name.rsplit("\\", 1)[-1],
-        unc_path=header.file_name,
+        unc_path=NOT_APPLICABLE,
```
slopez-b, line 90: Is this an expected change? We are replacing the path with `N/A`. This seems like something that should be in its own PR.

joshua-benchling: Thanks for catching up on this

PR #696, opened 2024-10-02
fix: SoftMax Pro - raise error for unsupported Group data format

slopez-b:
This error was reported in a sentry issue with only the traceback indicanting that a `ValueError` was raised.

Since there are no steps to reproduce but just the traceback I was unable to figure out the format in the `Group` data block that could result in this error, so no additional testing was added for this.

### DIFF
--- a/src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
+++ b/src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
@@ -203,13 +203,15 @@ class GroupData:
         )
 
         samples = data["Sample"].ffill()
-        return GroupData(
-            name=name,
-            sample_data=[
+        try:
+            sample_data = [
                 GroupSampleData.create(data.iloc[sample_entries.index])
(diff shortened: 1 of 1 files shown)


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -203,13 +203,15 @@ def create(reader: CsvReader) -> GroupData:
...
-            sample_data=[
+        try:
+            sample_data = [
                 GroupSampleData.create(data.iloc[sample_entries.index])
                 for _, sample_entries in samples.groupby(samples)
-            ],
-        )
+            ]
+        except ValueError as e:
+            msg = "Invalid data - unsupported Group data format."
```
nathan-stender, line 212: Can we add a bit more detail here: 

```suggestion
            msg = f"Unable to read Group data format for group {name}."
```

nathan-stender (changes requested):

nathan-stender (approved):

PR #721, opened 2024-10-11
feat: include unc path in all existing adapters

slopez-b:
Include the real file path and map it to the `data system document` `UNC_path` in all existing adapters.

- The `NamedFileContents` class will now get the full `file_path` instead of just the `file_name` of the instrument file being converted.
- This file path is now used for both the `file name` and `UNC path` properties of the `data system document` when they are not extracted from the instrument files.

TODO:
- Currently the tests are getting the absolute path from where they are stored, this will obviously make the tests fail when running in different systems. We need a way to mock the path to make the tests consistent. Probably the best place to do it will be when instantiating `NamedFileContets`


james-leinas (dismissed):
ASM output files LGTM 👍

nathan-stender (approved):

PR #728, opened 2024-10-18
fix: Fix error for NanoDrop Eight Adapter where the column names are different than the expected.

joshua-benchling:
- Standardize the column names used in the NanoDrop eight adapter


src/allotropy/parsers/thermo_fisher_nanodrop_eight/nanodrop_eight_reader.py
```diff
@@ -7,13 +7,22 @@
 from allotropy.parsers.lines_reader import CsvReader
 from allotropy.parsers.utils.pandas import read_csv
 
+COLUMNS_MAP = {
+    "Sample Name": ["Sample"],
```
nathan-stender, line 11: The preferred approach for this is to add options to the SeriesData getter, e.g:

`data.get(str, "Sample")` -> `data.get(str, ["Sample", "Sample Name"])`

This would address date/sample, but I noticed we only use "Sample" to detect the start of the table. 

I went ahead and refactored the reader a bit to make it more robust, and addressed this fix while I was at it, here: https://github.com/Benchling-Open-Source/allotropy/pull/731

So, you can go ahead and close this PR out.

nathan-stender (changes requested):

PR #743, opened 2024-10-24
feat: Novabio Flex2 - Update Adapter to use Solution Analyzer schema REC/2024/09

joshua-benchling:


src/allotropy/parsers/novabio_flex2/novabio_flex2_structure.py
```diff
@@ -165,6 +169,14 @@ def _get_measurements(sample: Sample) -> list[Measurement]:
             key: getattr(sample, key)
             for key in DETECTION_PROPERTY_MAPPING[detection_type]
         }
+        data_processing = {
+            key: kwargs.pop(key) for key in DATA_PROCESSING_FIELDS if key in kwargs
+        }
+        data_processing = dict(
+            filter(lambda item: item[1] is not None, data_processing.items())
+        )
```
slopez-b, line 177: You can include a condition in the first comprehension to do without the extra filtering:

```suggestion
        data_processing = {
            key: value for key in DATA_PROCESSING_FIELDS
            if key in kwargs and (value := kwargs.pop(key)) is not None
        }
```

slopez-b (changes requested):
Code LGTM, just a small suggestion

slopez-b (approved):

PR #745, opened 2024-10-25
fix: Agilent Gen5 Image - include transmission light setting when it is inside the channel setting

joshua-benchling:
Include transmission light setting when is set on the channel settings line.


src/allotropy/parsers/agilent_gen5_image/agilent_gen5_image_structure.py
```diff
@@ -89,8 +89,14 @@ class InstrumentSettings:
     @classmethod
     def create(cls, settings_lines: list[str]) -> InstrumentSettings:
         channel_settings = cls._get_channel_line_settings(settings_lines[0])
-
+        transmitted_light = None
+        if not channel_settings:
```
nathan-stender, line 93: Reading this line without context, it's confusing why we check `if not channel_settings`

Can you change the names of `channel_settings` / `_get_channel_line_settings` to something like `fluorescence_channel_settings` / `_get_fluorescence_settings`? 

That would make it a bit more clear that we get `transmitted_light` setting when the channel is not a fluorescence channel.

nathan-stender (changes requested):
Code mostly LGTM - one requested change to help make code behavior clearer

nathan-stender (approved):

PR #767, opened 2024-11-08
feat: Internal - add initial json-to-csv library

nathan-stender:
Adds `to_dataset.map_json` - a utility to map JSON data to a set of datasets and json dictionaries.

By providing a configuration file that specifies how JSON paths should be mapped to one or more datasets, along with some rudimentary transformations, this library allows users to construct custom datasets from ASM JSON results.


src/allotropy/json_to_csv/mapper_config.py
```diff
@@ -0,0 +1,308 @@
...
+class PivotTransformConfig(TransformConfig):
+    type_: TransformType
+    dataset: str
+    path: str
+    value_path: str
+    label_path: str
+
+    @staticmethod
+    def create(config_json: dict[str, Any]) -> PivotTransformConfig:
+        for key in ("dataset", "path", "value_path", "label_path"):
```
stephenworlow, line 50: nit: The key checking code between the JOIN and PIVOT seem identical. Could definitely pop out a required fields and abstract this away. 

It's only two examples though, so I'm not particular in how much is abstracted.

src/allotropy/json_to_csv/json_to_csv.py
```diff
@@ -0,0 +1,249 @@
...
+            single_values[column_config.name] = [value]
+    df = pd.DataFrame(single_values)
+
+    # Map nested dictionaries and lists, combining them with the base dataset.
+    for key, value in data.items():
+        path = Path(current_path, key)
+        path_df: pd.DataFrame = pd.DataFrame()
+
+        if isinstance(value, dict):
+            path_df = _map_dataset(value, config, path)
```
stephenworlow, line 107: nit: Given that current_path param is optional, can we be explicit?

src/allotropy/json_to_csv/json_to_csv.py
```diff
@@ -0,0 +1,249 @@
...
+        elif isinstance(value, list):
+            path_df = pd.concat(
+                [_map_dataset(item, config, path) for item in value],
```
stephenworlow, line 110: Same

src/allotropy/json_to_csv/json_to_csv.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def _rename_column(
+    df: pd.DataFrame, column_name: str
+) -> tuple[pd.DataFrame, list[str]]:
+    if column_name not in df:
+        return df, []
+
+    # Get column names to be used for rename.
+    labels = re.findall(r"\$([^\$]*)\$", column_name)
```
stephenworlow, line 140: This REGEX is used twice and is a bit arcane. Mind making it a constant with a description `r"\$([^\$]*)\$`

src/allotropy/json_to_csv/mapper_config.py
stephenworlow: Ah I see, not all of these Configs inherit from the same class. I'm 100% indifferent

stephenworlow (changes requested):

src/allotropy/json_to_csv/mapper_config.py
nathan-stender: I still want to do something about this - just not on the top of my list :D

src/allotropy/json_to_csv/json_to_csv.py
nathan-stender: I'll do you one better - it's only optional for the starting case - I remove the optional and had `map_dataset` call it with an empty path

src/allotropy/json_to_csv/json_to_csv.py
nathan-stender: Good point - I was being lazy by using the regex in json_to_csv, it should be contained to the column config. Fixed by adding `labels` property to column config.

src/allotropy/json_to_csv/mapper_config.py
nathan-stender: But fine I wont be lazy, fixed :P

stephenworlow (dismissed):

stephenworlow (dismissed):

stephenworlow (approved):

PR #768, opened 2024-11-08
feat: Agilent Gen5 - Support results not grouped in a single table 

slopez-b:
Gen5 adapters only supported measurements and calculated data when they were reported in a compiled "Results" matrix. For this, the `Regroup data in one matrix/table` option needed to be used in the configuration of the instrument.

This PR adds support for data reported in different named result tables.


james-leinas (approved):
ASM LGTM! 🎉

nathan-stender (approved):

PR #769, opened 2024-11-09
chore: Update to use license classifier instead of field

nathan-stender:
pypi seems to have stopped supporting the other way


stephenworlow (approved):
ooc, what does this affect?

Nevermind...I read the description

PR #770, opened 2024-11-12
feat: AppBio QuantStudio RT-PCR - handle format where only Results section is provided

nathan-stender:
Handles three alternative format cases:
1. Sample Setup section is not provided, metadata columns are in Results section instead
2. No Amplification Data section is provided - this data is optional, so treat as such
3. Block Type is not specified - use actual wells to determine plate_well_count instead


src/allotropy/parsers/appbio_quantstudio/appbio_quantstudio_data_creator.py
```diff
@@ -283,12 +284,39 @@ def _create_measurement_group(
...
+    well_location = [
+        well_item.position
+        for well in wells
+        for well_item in well.items
+        if well_item.position
+    ]
+    largest_column = sorted([str(loc[0]) for loc in well_location])[-1]
+    largest_row = sorted(int(loc[1:]) for loc in well_location)[-1]
+    well_number_by_position = (ord(largest_column.upper()) - ord("A") + 1) * largest_row
+    largest_well_nubmer = max(sorted(well_ids)[-1], well_number_by_position)
```
stephenworlow, line 308:
```suggestion
    largest_well_number = max(sorted(well_ids)[-1], well_number_by_position)
```

stephenworlow (approved):

PR #774, opened 2024-11-13
feat: Thermo Fisher Scientific VISIONlite - support new file format and calculated data

slopez-b:
- Support csv files separated by tabs
- Support Absorbance columns with a different name pattern (`A<Wavelength>`)
- Support files with `Result` column and add it to the calculated data documents


james-leinas (approved):
ASM looks great! 💯

nathan-stender (approved):

PR #777, opened 2024-11-18
chore: update liquid chromatography schema

Preston-O-Benchling:
benchling/2023/09


james-leinas (approved):

nathan-stender:
@Preston-O-Benchling it looks like the first two changes here were not signed by your GPG key, which will block merging. Can you try creating a new branch/PR? You can check that commits are signed by checking they have a green check by them in github. 

I'm not sure what happened, I know we set this up for you, and the third commit is signed (so you have signing set up) but the first two commits were not signed :-/

PR #782, opened 2024-11-21
feat: AppBio QuantStudio RT-PCR - support xlsx format

slopez-b:
Support XLSX format for AppBio Quantstudio and add extra logic for getting Cycle threshold data when a different Quantification cycle method is used (possible variations are `Ct`, `Crt`, `Cq`)


nathan-stender (changes requested):
Code mostly looks good - a couple requests

nathan-stender (approved):

james-leinas (approved):

PR #784, opened 2024-11-22
fix: Molecular Devices SoftMax Pro - forward-fill columns in group summaries correctly

nathan-stender:
Previously, a column in a group summary was assumed to be either:

1. Complete (every row has a value for the column), or 
2. Aggregated (every row except the first is blank, all rows share the same value)

This change updates to allow for the general case: empty values are forward-filled with the last filled value, but a column can have multiple values.

This change also creates `-0.0` entries for calculated values with errors, where previously they were skipped.

This is essentially a no-op for the ASM schema, but makes it easier to map datasets (since the -0 value tells us there is a calculated value)


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_data_creator.py
```diff
@@ -204,40 +204,21 @@ def _get_reduced_calc_docs(data: StructureData) -> list[CalculatedDataItem]:
     ]
 
 
-def _get_group_agg_calc_docs(
```
nathan-stender, line 207: Note to reviewer: this is the special case for aggregated rows, removed in favor of generalizing the other case.

tests/parsers/moldev_softmax_pro/testdata/MD_SMP_absorbance_endpoint_example01.json
nathan-stender, line 1: Note to reviewer: the updates to existing files are just re-ordering of calculated data documents, because previously, "aggregated" calc values were added separately after the "normal" values.

PR #786, opened 2024-11-22
feat: MSD Discovery Workbench - initial implementation

joshua-benchling:
- Added new adapter called msd workbench
- updated plate reader schema mapper to handle new adapter


james-leinas (dismissed):
ASM Output LGTM! 🎉

src/allotropy/parsers/msd_workbench/msd_workbench_parser.py
```diff
@@ -0,0 +1,52 @@
...
+    Header,
+    PlateData,
+)
+from allotropy.parsers.release_state import ReleaseState
+from allotropy.parsers.vendor_parser import VendorParser
+
+
+class MSDWorkbenchParser(VendorParser[Data, Model]):
+    DISPLAY_NAME = "MSD Discovery Workbench"
+    RELEASE_STATE = ReleaseState.WORKING_DRAFT
```
james-leinas, line 29: @joshua-benchling, can you update RELEASE_STATE to "Recommended".

james-leinas (dismissed):

src/allotropy/parsers/msd_workbench/msd_workbench_structure.py
```diff
@@ -0,0 +1,168 @@
...
+    LUMINESCENCE_DETECTOR,
+    SAMPLE_ROLE_TYPE_MAPPING,
+    SOFTWARE_NAME,
+)
+from allotropy.parsers.utils.pandas import SeriesData
+from allotropy.parsers.utils.uuids import random_uuid_str
+
+
+@dataclass(frozen=True)
+class Header:
```
nathan-stender, line 30: I think it's cleaner if you just remove this class and pass `file_name` into `create_metadata`

src/allotropy/parsers/msd_workbench/msd_workbench_structure.py
```diff
@@ -0,0 +1,168 @@
...
+        if "Plate" not in first_row:
+            msg = "Plate ID not found in the first row of the data"
+            raise AllotropeConversionError(msg)
+        well_plate_id = first_row.split("_")[-1].strip()
+        data = data.iloc[1:].reset_index(drop=True)
+        # Set the first row as the header
+        data.columns = pd.Index(data.iloc[0])
+        data = data[1:].reset_index(drop=True)
+        well_data = []
+        for _row_index, row in data.iterrows():
```
nathan-stender, line 65: `WellData.create` should take in the `SeriesData` 

Then you can replace this for loop with `map_rows(data, WellData.create)`

src/allotropy/parsers/msd_workbench/msd_workbench_structure.py
```diff
@@ -0,0 +1,168 @@
...
+                    measurement_custom_info={
+                        "detection range": row_series.get(str, "Detection Range"),
+                        "assay identifier": row_series.get(str, "Assay"),
+                    },
+                    dilution_factor=row_series.get(int, "Dilution Factor"),
+                )
+            )
+
+        well_column = data["Well"]
+        plate_well_count = len(well_column.unique())
```
nathan-stender, line 82: We should not rely on the raw count here for the plate well count, we should "round up" to a plate well count that we know it correct. See `_get_plate_well_count` in `appbio_quantstudio_data_creator`. You can use similar logic.

It's fine if you even just copy that logic but adapter it to pass in the well ids instead of the `Well` dataclass used there

src/allotropy/parsers/msd_workbench/msd_workbench_structure.py
```diff
@@ -0,0 +1,168 @@
...
+    ) -> PlateData:
+        first_row = str(data.iloc[0, 0])
+        if "Plate" not in first_row:
+            msg = "Plate ID not found in the first row of the data"
+            raise AllotropeConversionError(msg)
+        well_plate_id = first_row.split("_")[-1].strip()
+        data = data.iloc[1:].reset_index(drop=True)
+        # Set the first row as the header
+        data.columns = pd.Index(data.iloc[0])
+        data = data[1:].reset_index(drop=True)
```
nathan-stender, line 63: The above should be done in the reader class. You can store the first line as a separate string, drop it, and then read the CSV into a dataframe, without having to do this manipulation.

Then here, you can just read the well_plate_id from that first line, and use the df as is

src/allotropy/parsers/msd_workbench/msd_workbench_structure.py
```diff
@@ -0,0 +1,168 @@
...
+        grouped_wells[well.sample_identifier].append(well)
+
+    well_data: list[list[WellData]] = list(grouped_wells.values())
+
+    return [
+        MeasurementGroup(
+            measurement_time=plate_data.measurement_time,
+            plate_well_count=plate_data.plate_well_count,
+            measurements=[
+                Measurement(
```
nathan-stender, line 148: This seems like a pretty 1:1 mapping from fields in "Well" to the measurement fields, I think should should probably just create the measurements directly from rows (no need for intermediate classes - these are really only needed when there are multiple different sections of data that needs to be used to create a measurement)

nathan-stender (changes requested):

nathan-stender (approved):

james-leinas (dismissed):

nathan-stender (approved):

PR #795, opened 2024-11-27
fix: MSD Methodical Mind - handle case were there are empty identifiers, and handle digital signature at the top of the file

joshua-benchling:
- Skipped Digital Signature top line
- Added unique index identifiers list


james-leinas (changes requested):
@joshua-benchling, the `plate well count` field looks to be recorded as "108". Expected value should be 96. 

We should implement the recording based on this in the requirements:
image

james-leinas (changes requested):
@joshua-benchling, also looks like the `location identifier` field records the plate row (e.g. "A", "B", "C") only for the first row "A". Rest of the file contains only the plate column. Potentially formatting difference within the _Data_ (Plate Matrix)?

james-leinas (approved):

nathan-stender (approved):
Code LGTM - I changed the title - please check to make sure I got it right.  Going forward, please make sure the description of the fix is always a complete sentence that describes the change (remember these become changelog entries!)

PR #798, opened 2024-12-03
feat: Tecan Magellan - Initial implementation

slopez-b:
Add Tecan Magellan plate reader adapter. This adapter supports only Absorbance measurements at the moment. More measurement modes support will be added when there's available example data.


src/allotropy/parsers/tecan_magellan/tecan_magellan_structure.py
```diff
@@ -0,0 +1,204 @@
...
+
+def get_measurement_type(measurement_mode: str) -> MeasurementType:
+    if measurement_mode.lower() == "absorbance":
+        return MeasurementType.ULTRAVIOLET_ABSORBANCE
+
+    msg = f"Only Absorbance measurements are supported as this time. Got {measurement_mode}."
+    raise AllotropeConversionError(msg)
+
+
+def parse_settings_lines(lines: list[str]) -> SeriesData:
```
nathan-stender, line 33: Could we do this in the reader class? So the reader would end up with a metadata or settings SeriesData memeber.

src/allotropy/parsers/tecan_magellan/tecan_magellan_structure.py
```diff
@@ -0,0 +1,204 @@
...
+        software_name=constants.SOFTWARE_NAME,
+        product_manufacturer=constants.PRODUCT_MANUFACTURER,
+        equipment_serial_number=data.equipment_serial_number,
+    )
+
+
+def create_measurement_groups(
+    data: pd.DataFrame, metadata: MagellanMetadata
+) -> list[MeasurementGroup]:
+    return [
```
nathan-stender, line 162: You can use `functools.partial` to bind the common aguments of the function and use map_rows. This might not be exactly correct, but I think it's pretty close:

```suggestion
    from functools import partial
    return [group for group in map_rows(data, partial(_create_measurement_group, metadata=metadata, well_count=len(data))) if group]
```

src/allotropy/parsers/utils/pandas.py
```diff
@@ -121,11 +121,13 @@ def split_header_and_data(
 def split_dataframe(
     df: pd.DataFrame,
     should_split_on_row: Callable[[pd.Series[Any]], bool],
+    include_split_row: bool = False,  # noqa: FBT001, FBT002
```
nathan-stender, line 124: To get rid of these qa warnings, put a * in arguments (this forces include_split_row to be kwarg only)
```suggestion
    *, include_split_row: bool = False,
```

src/allotropy/parsers/tecan_magellan/tecan_magellan_structure.py
slopez-b: Thanks for the suggestion!

I actually ended up adding logic for capturing missing measurements to error documents and got rid of the condition at the end.

src/allotropy/parsers/tecan_magellan/tecan_magellan_structure.py
nathan-stender: Discussed offline - missed that you are parsing settings lines out of multiple sections.

nathan-stender (approved):

james-leinas (approved):

PR #805, opened 2024-12-06
fix: Global - update schemas to align the use of `μ` characters with allotrope

slopez-b:
There are two characters with the same symbol to represent `micro` units. Allotrope uses a mixture of those symbols in its units definitions. This was previously handled by standardizing the use of the same `μ` character in all `allotropy` schemas. 
However, this might present a maintainability problem as we'd have to search and replace the character every time we add a new schema. Another drawback is that the `ASM` we generate would probably fail validation against the original Allotrope schemas.
This PR reconciles the characters used in our schemas with the source Allotrope units schema.


nathan-stender (approved):
Thanks for taking care of this 🎉

PR #806, opened 2024-12-06
feat: Unchained Labs Lunatic - update adapter to map to Plate Reader REC/2024/06 schema

joshua-benchling:
- Updated adapter to work with Plate Reader 24/06 schema mapper.
- Added error documents for missing absorbance and calculated data values.


src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
```diff
@@ -63,40 +64,64 @@ def _create_measurement(
...
         type_=MeasurementType.ULTRAVIOLET_ABSORBANCE,
         device_type=DEVICE_TYPE,
         detection_type=DETECTION_TYPE,
         identifier=measurement_identifier,
         detector_wavelength_setting=float(wavelength),
         electronic_absorbance_reference_wavelength_setting=background_wavelength,
-        electronic_absorbance_reference_absorbance=background_absorbance,
-        absorbance=well_plate_data.get(float, wavelength_column, NaN),
+        measurement_custom_info={
+            "electronic_absorbance_reference_absorbance": background_absorbance
```
nathan-stender, line 90:
```suggestion
            "electronic absorbance reference absorbance": background_absorbance
```

Also - it appears that `measurement_custom_info` is not being used in the schema mapper - not sure why it was added to the dataclass but not used - but you'll want to add the code to use it so this doesn't get dropped

src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
```diff
@@ -63,40 +64,64 @@ def _create_measurement(
...
+        },
+        absorbance=absorbance if absorbance is not None else NEGATIVE_ZERO,
         sample_identifier=well_plate_data[str, "Sample name"],
         location_identifier=well_plate_data[str, "Plate Position"],
         well_plate_identifier=well_plate_data.get(str, "Plate ID"),
         batch_identifier=well_plate_data.get(str, "Sample Group"),
         firmware_version=header.get(str, "Client version"),
-        path_length=float(path_length) if path_length else None,
+        sample_custom_info={
+            "path length": path_length if path_length else None,
```
nathan-stender, line 99: Let's still cast this to a float

```suggestion
            "path length": float(path_length) if path_length is not None else None,
```

nathan-stender (changes requested):

src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
joshua-benchling: It is being used in the `_get_luminescence_measurement_document`. Other technique measurement documents could use when needed.

tests/parsers/unchained_labs_lunatic/testdata/spectrum_measurement.json
```diff
@@ -81,13 +73,9 @@
                             "absorbance": {
                                 "value": 0.696,
                                 "unit": "mAU"
-                            },
-                            "custom information document": {
```
nathan-stender, line 85: I bring up custom info doc because I see this is being dropped

nathan-stender (changes requested):

src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
nathan-stender: See comment below about value that is being removed

nathan-stender (approved):

PR #819, opened 2024-12-12
feat: Molecular Devices SoftMax Pro - add support for spectrum data

joshua-benchling:
- Added support to reading and parsing spectrum data
- Updated plate reader schema mapper to support custom info doc at measurement group level


tests/parsers/moldev_softmax_pro/testdata/spectrum_data.txt
```diff
@@ -0,0 +1,2351 @@
+##BLOCKS= 97
```
nathan-stender, line 1: Lets cut down this test file size to 2 or 3 wavelengths, to keep test time down.

tests/parsers/moldev_softmax_pro/to_allotrope_test.py
```diff
@@ -53,17 +53,3 @@ def test_invalid_measurement_table_dimensions() -> None:
         match="Expected 4 rows in measurement table, got 3.",
     ):
         from_file(f"{TESTDATA}/errors/missing_measurement_table_row.txt", VENDOR_TYPE)
-
-
-@pytest.mark.parametrize(
-    "test_file",
-    [
-        f"{TESTDATA}/errors/lum_spectrum_columns.txt",
```
nathan-stender, line 61: It looks like this file still does not work with our parser.

Please add logic to detect this and raise an error message.

And example could be detecting if the read_type == Spectrum and the measurment type == lum, raising an error saying we don't support lum spectrum format at this time.

nathan-stender (changes requested):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -437,28 +437,6 @@ def create(reader: CsvReader, header: PlateHeader) -> PlateRawData:
 
         return plate_raw_data
 
-    @staticmethod
```
nathan-stender, line 440: I see why you moved this, but moving it to the `PlateData` class isn't ideal either, as is break encapsulation.

I would suggest either:

1. Keeping 1 `RawPlateData` class, and having `create` do the different logic based on read_type
2. Having a base class `RawPlateData` that is a base class to two subclasses `SpectrumRawPlateData` and something like `EndpointAndKineticRawPlateData`, then keeping common logic (like this function) in the base class

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -819,6 +896,9 @@ def parse_header(cls, header: pd.Series[str]) -> PlateHeader:
         cls.check_export_version(export_version)
         cls.check_read_type(read_type)
         cls.check_data_type(data_type)
+        if read_type == ReadType.SPECTRUM.value:
+            msg = "Spectrum read type is not supported for fluorescence plates."
```
nathan-stender, line 900:
```suggestion
            msg = "Spectrum read type is not currently supported for fluorescence plates."
```

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -934,6 +1014,9 @@ def parse_header(cls, header: pd.Series[str]) -> PlateHeader:
         cls.check_export_version(export_version)
         cls.check_read_type(read_type)
         cls.check_data_type(data_type)
+        if read_type == ReadType.SPECTRUM.value:
+            msg = "Spectrum read type is not supported for luminescence plates."
```
nathan-stender, line 1018:
```suggestion
            msg = "Spectrum read type is not currently supported for luminescence plates."
```

tests/parsers/moldev_softmax_pro/to_allotrope_test.py
```diff
@@ -64,6 +64,6 @@ def test_invalid_measurement_table_dimensions() -> None:
 def test_unrecognized_read_type(test_file: str) -> None:
     with pytest.raises(
         AllotropeConversionError,
-        match="Only Endpoint or Kinetic measurements can be processed at this time.",
+        match="Spectrum read type is not supported for luminescence plates.",
```
nathan-stender, line 67:
```suggestion
        match="Spectrum read type is not currently supported for luminescence plates.",
```

nathan-stender (changes requested):
One last tiny language change request

Also - please change the parser name to `Molecular Devices SoftMax Pro` in the title

james-leinas (approved):
ASM output 👍

PR #823, opened 2024-12-13
feat: Cytiva Unicorn - allow optional values

alejandro-salgado:
- redefine some static docs attributes as optional in unicorn adapter 
- add optional methods to strict xml element 
- allow optional zip extension in unicorn adapter


james-leinas (approved):
Tested additional zip and produced expected ASM! 🎉

nathan-stender (approved):

PR #831, opened 2024-12-27
feat: BMG Labtech SMART Control - initial implementation

joshua-benchling:
Added new adapter Bmg Labtech Smart Control


tests/parsers/bmg_labtech_smart_control/testdata/fluroescence_endpoint_example1.json
```diff
@@ -0,0 +1,5340 @@
...
+                            },
+                            "measurement identifier": "BMG_LABTECH_SMART_CONTROL_TEST_ID_0",
+                            "sample document": {
+                                "sample identifier": "Standard S1",
+                                "location identifier": "A01",
+                                "sample role type": "standard sample role",
+                                "well plate identifier": "Plate 3",
+                                "mass concentration": {
+                                    "value": 1000.0,
+                                    "unit": "pg/mL"
```
james-leinas, line 51: @joshua-benchling, since SMART Control reports the concentration in ng/mL and the Plate Reader schema accepts only pg/mL, we'll need to do a conversion from ng/mL within the adapter.

image

tests/parsers/bmg_labtech_smart_control/testdata/fluroescence_endpoint_example1.json
```diff
@@ -0,0 +1,5340 @@
...
+                    "measurement document": [
+                        {
+                            "device control aggregate document": {
+                                "device control document": [
+                                    {
+                                        "device type": "fluorescence detector",
+                                        "detection type": "fluorescence",
+                                        "detector distance setting (plate reader)": {
+                                            "value": 5.9,
+
```
james-leinas, line 5318: @joshua-benchling, it looks like the data source identifier's for the `fluorescence` and `Average of all blanks used` features are swapped. The `fluorescence` feature should align to the first measurement document's `measurement identifier`

james-leinas (changes requested):

james-leinas (dismissed):
🎉

src/allotropy/parsers/bmg_labtech_smart_control/bmg_labtech_smart_control_structure.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def _get_measurement_time(date: str, time: str) -> str:
+    datetime_str = f"{date} {time}"
+    datetime_obj = datetime.strptime(datetime_str, "%m/%d/%Y %I:%M:%S %p").astimezone()
+    formatted_datetime = datetime_obj.strftime("%Y-%m-%d %H:%M:%S")
+    return formatted_datetime
+
+
+def _get_target_temperature(row: SeriesData) -> float | None:
```
nathan-stender, line 149: Nit, but I'd just move this logic inside the function, we already do simple 1 or 2 line processing on other fields.

In fact, because "set off" won't parse to a float, and you're just setting to None anyway, you can replace this with just: `row.get(float, TARGET_TEMPERATURE)` and it will work the same.

src/allotropy/parsers/bmg_labtech_smart_control/bmg_labtech_smart_control_structure.py
```diff
@@ -0,0 +1,217 @@
...
+        return []
+
+    non_blank_measurements = [
+        measurement
+        for group in measurement_groups
+        for measurement in group.measurements
+        if measurement.sample_role_type != SampleRoleType.blank_role
+    ]
+
+    blank_average = sum(
```
nathan-stender, line 174: We should not be manually calculating calculated data values, we should only report what is in the file.

I checked with @james-leinas and he says that this value should be the file

nathan-stender (changes requested):

src/allotropy/parsers/bmg_labtech_smart_control/bmg_labtech_smart_control_parser.py
```diff
@@ -0,0 +1,35 @@
...
+
+class BmgLabtechSmartControlParser(VendorParser[Data, Model]):
+    DISPLAY_NAME = DISPLAY_NAME
+    RELEASE_STATE = ReleaseState.RECOMMENDED
+    SUPPORTED_EXTENSIONS = BmgLabtechSmartControlReader.SUPPORTED_EXTENSIONS
+    SCHEMA_MAPPER = Mapper
+
+    def create_data(self, named_file_contents: NamedFileContents) -> Data:
+        reader = BmgLabtechSmartControlReader(named_file_contents)
+        measurement_groups = create_measurement_groups(reader.header, reader.data)
```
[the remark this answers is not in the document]
joshua-benchling: But this is actually used twice, as an argument for `Data` and an argument for `create_calculated_data_documents`

src/allotropy/parsers/bmg_labtech_smart_control/bmg_labtech_smart_control_structure.py
```diff
@@ -0,0 +1,217 @@
...
+        file_name=path.name,
+        unc_path=file_path,
+        software_name=SOFTWARE_NAME,
+    )
+
+
+def create_measurement_groups(
+    headers: SeriesData, data: pd.DataFrame
+) -> list[MeasurementGroup]:
+    def map_measurement_group(row: SeriesData) -> MeasurementGroup:
```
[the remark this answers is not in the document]
joshua-benchling: but I did use a global state outside of the function, which is the `headers`. The thing is that `map_rows` can´t pass additional arguments besides the data and the handler.

james-leinas (commented):
@nathan-stender, asked @joshua-benchling to log a warning rather than fail if we are unable to extract the average of blanks value from the file, since calculated data documents are optional. Will help to debug if issue with extracting the data

src/allotropy/parsers/bmg_labtech_smart_control/bmg_labtech_smart_control_parser.py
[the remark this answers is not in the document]
nathan-stender: Oops, I see 👍

src/allotropy/parsers/bmg_labtech_smart_control/bmg_labtech_smart_control_structure.py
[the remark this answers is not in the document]
nathan-stender: Ah, I didn't see that you had to use another argument. There actually is a way to do this: functools.partial

You can make:

`create_measurement_group(data: SeriesData, headers: SeriesData)`

and then call

`map_rows(data, partial(create_measurement_group, headers=headers))`

See tecan_magellan_parser for an example!

nathan-stender (changes requested):
Thanks for the fixes, and sorry I missed a couple things in my review - one more change!

james-leinas (dismissed):

src/allotropy/parsers/bmg_labtech_smart_control/constants.py
```diff
@@ -0,0 +1,13 @@
+from allotropy.allotrope.models.shared.components.plate_reader import SampleRoleType
+
+DISPLAY_NAME = "BMG SMART Control"
```
nathan-stender, line 3:
```suggestion
DISPLAY_NAME = "BMG Labtech SMART Control"
```
Sorry just noticed this

nathan-stender: (will need to update auto-generated tables as well)

nathan-stender (approved):

PR #833, opened 2025-01-02
feat:  AppBio QuantStudio RT-PCR - Use qpcr rec/2024/09 schema mapper

slopez-b:
- Added schema mapper for qpcr/REC/2024/09 schema
- Migrated Appbio Quantstudio adapter to the new schema


tests/parsers/appbio_quantstudio/testdata/appbio_quantstudio_minimal_test02.json
```diff
@@ -1,57 +1,48 @@
...
+                                    "target color": "RGB(176,23,31)"
+                                }
+                            },
                             "processed data aggregate document": {
                                 "processed data document": [
                                     {
                                         "data processing document": {
-                                            "cycle threshold value setting": {
+                                            "automatic cycle threshold enabled setting": true,
+                                            "cycle threshold value setting  qPCR ": {
```
stephenworlow, line 42: It's a bit odd that are two spaces between setting and qPRC. Is that intended?

stephenworlow (dismissed):
One small question about the formatting of one of the schema keys, but otherwise LGTM

stephenworlow:
Still have some failing tests, so I'll wait for them before approving!

tests/parsers/appbio_quantstudio/testdata/appbio_quantstudio_minimal_test02.json
slopez-b: looking into it.

slopez-b:
Note to reviewer:

I had to update some of the `appbio quantstudio design and analysis` tests because that adapter uses the datacube creation classes modified in this PR. These changes don't compromise the actual contents of the datacubes but the dimensions types being `float` now instead of `int`.

tests/parsers/appbio_quantstudio/testdata/appbio_quantstudio_minimal_test02.json
slopez-b: Great catch! this has to do to the way we parse keys in the schemas that include parenthesis. It seems like we have to add those cases to a `SPECIAL_KEYS` dictionary in the converter script. Working on it.

stephenworlow (approved):

PR #835, opened 2025-01-03
feat: MSD Workbench - add .txt file support

joshua-benchling:
* Added methodical mind call from msd adapter so it handles .txt files
* Added methodical mind test file to msd workbench test cases


nathan-stender (changes requested):
Code LGTM! 

It looks like you forgot to add the `methodical_test_1.txt` file, (the test doesn't fail because it simply doesn't run the test case, even though you added the expected result file) - can you please add that?

james-leinas (approved):

nathan-stender (approved):

PR #837, opened 2025-01-07
feat: Bio-Rad CFX Maestro - Use qpcr rec/2024/09 schema mapper

slopez-b:


james-leinas (dismissed):

src/allotropy/parsers/cfxmaestro/cfxmaestro_structure.py
```diff
@@ -68,11 +78,19 @@ def create_measurement_group(
...
                         float, "Cq", validate=SeriesData.NOT_NAN
                     ),
-                    # TODO: confirm the exported column name for cycle number
+                    # TODO: cycle number is required, but we do not have any examples of the value being
+                    # provided, if we get one, and the column does not match, update here.
                     cycle_threshold_value_setting=data.get(
                         float, "Cycle Number", NEGATIVE_ZERO
                     ),
                 ),
+                # Since the processed data doc does not include an error document,
```
nathan-stender, line 87: Because we are trying to get the value, we should only report this if `data.get(float, "Cycle Number", NEGATIVE_ZERO)` == `NEGATIVE_ZERO`

src/allotropy/parsers/cfxmaestro/cfxmaestro_structure.py
```diff
@@ -68,11 +78,19 @@ def create_measurement_group(
...
+                # we added the cycle threshold value setting error at measurement level
+                error_document=[
+                    Error(
+                        error=NaN.value, feature="cycle threshold value setting (qPCR)"
```
nathan-stender, line 91: Can we update this error to "Value not provided"?

src/allotropy/parsers/cfxmaestro/cfxmaestro_structure.py
```diff
@@ -58,7 +65,10 @@ def create_measurement_group(
                 ),
                 timestamp=DEFAULT_EPOCH_TIMESTAMP,
                 # Optional measurement metadata
-                sample_role_type=data.get(str, "Content"),
+                sample_role_type=constants.SAMPLE_ROLE_TYPES_MAP.get(
+                    data.get(str, "Content", "__INVALID_KEY__")
```
nathan-stender, line 69: @james-leinas for unexpected value, should we put `unknown` sample type?

src/allotropy/parsers/cfxmaestro/cfxmaestro_structure.py
```diff
@@ -43,6 +47,9 @@ def create_measurement_group(
 ) -> MeasurementGroup:
     return MeasurementGroup(
         plate_well_count=plate_well_count,
+        well_volume=NEGATIVE_ZERO,
+        error_document=[Error(error=NaN.value, feature="well volume")],
```
nathan-stender, line 51: Can we change error here to "Value not provided"? Thoughts @james-leinas ?

My concern is that `NaN` == `Not a number` which doesn't quite describe the error here.

nathan-stender (changes requested):

src/allotropy/parsers/cfxmaestro/cfxmaestro_structure.py
james-leinas: @nathan-stender, good call there. Would agree that we put error instead as "Value not provided in instrument file"

src/allotropy/parsers/cfxmaestro/cfxmaestro_structure.py
james-leinas: @nathan-stender, unknown in context of some experiments may relate to unknown concentration and experiment is intended to determine. Since sample role type optional would lean to not report value instead if blank or an unexpected value

james-leinas (dismissed):

nathan-stender (changes requested):
Looks like there's a lint issue due to a typo in an import? 

Requesting changes to clear my review request - feel free to re-request once tests are passing.

nathan-stender (approved):

PR #843, opened 2025-01-13
chore: Updated LC Schema and Change Notes

Preston-O-Benchling:
Added Peak Index Field, Start Time Setting Field. 
Removed "Is Required" from Autosampler Injection Volume Setting field, Injection Time field, and UUID under Injection Document. 
Updated CHANGE_NOTES.md to reflect this.


james-leinas (commented):
@Preston-O-Benchling, think you just need to change title so "Chore" is not capitalized. Tests should pass then

james-leinas (approved):

PR #845, opened 2025-01-13
feat: Cytiva Unicorn - improve way to filter xml metadata

alejandro-salgado:
- improve way to filter xml metadata


alejandro-salgado:
already implemented in https://github.com/Benchling-Open-Source/allotropy/pull/841

PR #854, opened 2025-01-21
feat: Thermo Fisher Qubit4 - Add unread data to asm.

joshua-benchling:
* Added unread data (if any) to asm as custom doc.


src/allotropy/parsers/thermo_fisher_qubit4/thermo_fisher_qubit4_structure.py
```diff
@@ -84,6 +84,10 @@ def create_measurement_group(data: SeriesData) -> MeasurementGroup:
             )
         ],
     )
+    unread_data = data.get_unread()
```
nathan-stender, line 87: Please just move this into the measurement, keeping the dataclass frozen.

I think I see why you did this (in case there are multiple measurements) but there is only one measurement hard-coded into this group. If there were more, the code would change.

nathan-stender (changes requested):

nathan-stender (approved):

PR #855, opened 2025-01-21
feat: Thermo Fisher Qubit Flex - Add unread data to asm

joshua-benchling:


tests/parsers/thermo_fisher_qubit_flex/testdata/thermo_fisher_qubit_flex_example_02.json
```diff
@@ -64,6 +64,9 @@
                             "fluorescence": {
                                 "value": 156.0,
                                 "unit": "RFU"
+                            },
+                            "custom information document": {
+                                "Last Read Standards": "2023-02-11 10:47:55"
```
joshua-benchling, line 69: should this be lower case?

src/allotropy/parsers/thermo_fisher_qubit_flex/thermo_fisher_qubit_flex_structure.py
```diff
@@ -96,21 +92,25 @@ def create_measurement_group(data: SeriesData) -> MeasurementGroup:
...
             )
         ],
     )
 
 
 def create_data(df: pd.DataFrame, file_path: str) -> Data:
     header = df_to_series_data(df.head(1))
+    software_version = header.get(str, "Software Version")
+    header.get_custom_keys("Sample Volume \\(uL\\)")
+    header.get_unread()
```
nathan-stender, line 105: I generally agree with this approach of calling `get_unread` to discard unread keys for the header. 

However, can you put a comment explaining this - for those less familiar
```suggestion
    # We reader header info from a row in the table, so we don't need to read all keys from this SeriesData.
    # Call get_unread() to mark all keys as read and suppress the warning for unread keys.
    header.get_unread()
```

src/allotropy/parsers/thermo_fisher_qubit_flex/thermo_fisher_qubit_flex_structure.py
```diff
@@ -96,21 +92,25 @@ def create_measurement_group(data: SeriesData) -> MeasurementGroup:
...
+                custom_info=data.get_unread(skip={"Software Version"}),
             )
         ],
     )
 
 
 def create_data(df: pd.DataFrame, file_path: str) -> Data:
     header = df_to_series_data(df.head(1))
+    software_version = header.get(str, "Software Version")
+    header.get_custom_keys("Sample Volume \\(uL\\)")
```
nathan-stender, line 104: Is there some reason we need this call specifically? Shouldn't `get_unread` cover it?

nathan-stender (changes requested):

src/allotropy/parsers/thermo_fisher_qubit_flex/thermo_fisher_qubit_flex_structure.py
```diff
@@ -50,10 +50,6 @@ def create_measurement_group(data: SeriesData) -> MeasurementGroup:
                 original_sample_concentration_unit=data.get(
                     str, "Original sample conc. units"
                 ),
-                custom_info={
-                    "reagent lot number": data.get(int, "Reagent Lot#"),
-                    "calibrated tubes": data.get(int, "Calibrated Tubes"),
-                },
                 sample_custom_info={
                     "last read standards": data.get(str, "Test Date"),
```
james-leinas, line 54: @joshua-benchling, it looks like from our initial implementation `last read standards` is being mapped from `Test Date` rather than the specified column in the instrument file. Agree with its positioning in the hierarchy but would suggest to change column to map to as 'Last Read Standards'.

james-leinas (changes requested):

tests/parsers/thermo_fisher_qubit_flex/testdata/thermo_fisher_qubit_flex_example_01.json
```diff
@@ -42,7 +42,7 @@
                                 "sample identifier": "S1",
                                 "batch identifier": "131222-164034",
                                 "custom information document": {
-                                    "last read standards": "2022-12-13T16:41:09+00:00",
+                                    "Last Read Standards": "13/12/2022 04:41:09 PM",
```
nathan-stender, line 45: This data changed format because you changed the key - it needs to change in the schema mapper too

src/allotropy/parsers/thermo_fisher_qubit_flex/thermo_fisher_qubit_flex_structure.py
nathan-stender: I'm not sure this is what @james-leinas was asking for? 

I think he was saying to read "Last Read Standards" from the data, rather than "Test Date"?

@james-leinas to confirm

james-leinas: @nathan-stender yes, you got it. We are currently reading "Last Read Standards" from the Test Date column. Should swap it to read from "Last Read Standards". 

"Test Date" is already mapped to "measurement time"

joshua-benchling: My bad, I misunderstood that.
I mapped it to the right field now, thanks both.

nathan-stender (approved):

james-leinas (approved):

PR #880, opened 2025-02-07
feat: add void volume and flow rate to Cytiva Unicorn adapter

alejandro-salgado:
- add column volume and flow rate to Cytiva Unicorn adapter
- add column volume and flow rate to Liquid chromatography benchling 09/23 schema and mapper


src/allotropy/allotrope/schemas/adm/liquid-chromatography/BENCHLING/2023/09/liquid-chromatography.schema.json
```diff
@@ -117,6 +117,18 @@
                                               "$asm.pattern": "value datum",
                                               "$asm.type": "http://www.w3.org/2001/XMLSchema#string",
                                               "$ref": "http://purl.allotrope.org/json-schemas/adm/core/REC/2023/09/core.schema#/$defs/tStringValue"
+                                            },
+                                            "column volume": {
```
Preston-O-Benchling, line 121: Could this be changed from "column volume" to "void volume"? The property class is already defined as "void volume". The rest of the documents would need "column volume" switched to "void volume"

james-leinas (approved):

nathan-stender (approved):

PR #890, opened 2025-02-13
feat: rename conductivity and flow measurements in Cytiva Unicorn adapter

alejandro-salgado:
- rename electric conductivity to differentiate S/m from % measurements
- rename sample flow to differentiate CV/h, mL/min and cm/s
- rename system flow to differentiate CV/h and mL/min


james-leinas (commented):
@alejandro-salgado, we should probably keep the concept labels consistent with the same term we have currently. Understand the idea to be prescriptive to what the unit is but since we have a unit attached to each datacube, don't think we need to relabel the concepts.

PR #900, opened 2025-02-26
fix: MSD Methodical Mind - add type verification for luminescence

alejandro-salgado:
- redefine luminescence as float and add type verification in methodical mind


src/allotropy/parsers/methodical_mind/methodical_mind_structure.py
```diff
@@ -56,7 +57,7 @@ def create(
         ]
         well_data = [
             WellData.create(
-                luminescence=value,
+                luminescence=try_int(value, "luminescence"),
```
james-leinas, line 60: @alejandro-salgado, would you be able to change this to `try_float` instead? Most cases will have LUM reported as an int but some instruments can have more precision to report decimals

james-leinas (changes requested):

nathan-stender (approved):

stephenworlow (approved):

PR #918, opened 2025-03-06
fix: ChemoMetec Nucleoview - add validation for cell viability field.

joshua-benchling:
output example when cell viability not present in the file
image


james-leinas (dismissed):

src/allotropy/parsers/chemometec_nucleoview/nucleoview_reader.py
```diff
@@ -17,7 +17,6 @@ def read(cls, contents: IOType) -> pd.DataFrame:
             {"Estimated cell diameter [um]": "Estimated cell diameter (um)"},
             axis="columns",
         )
-        # Drop rows that do not report cell viability.
```
nathan-stender, line 20: This is not producing the same behavior as before.

Before, we would drop rows that do not have Viability and still succeed overall.

Now, if any row is missing Viability, we will fail overall.

Instead, we should just add a check here that is Viability column is missing, we raise and AllotropeConversionError

nathan-stender (changes requested):

src/allotropy/parsers/chemometec_nucleoview/nucleoview_reader.py
```diff
@@ -17,7 +18,9 @@ def read(cls, contents: IOType) -> pd.DataFrame:
...
-        df = df.dropna(subset=["Viability (%)"])
+        if "Viability (%)" not in df.columns:
+            msg = "Value for Cell Viability was expected, but was not found in the input file"
+            raise AllotropeConversionError(msg)
+        df = df.dropna()
```
nathan-stender, line 24:
```suggestion
        df = df.dropna(subset=["Viability (%)"])
```

nathan-stender (changes requested):

nathan-stender (approved):

PR #934, opened 2025-03-19
feat: Appbio Quantstudio design and analysis - redefine cycle threshold result as required data source

alejandro-salgado:
- add possibility to define a data source as required in calcdocs library
- redefine cycle threshold result as required data source in Appbio Quantstudio design and analysis


james-leinas (dismissed):
👍 -- additional feature support to be added for quantity values of STANDARD samples

nathan-stender (changes requested):
Code looks good - but we should really have a test that exercises this, even if we have to create a small one.

Can we take an existing small test and set it up so that this filter is exercised?

nathan-stender (approved):

PR #952, opened 2025-04-02
fix: Internal - improve error messaging for errors in various parsers

joshua-benchling:
* Added error handling for nova bioflex2 missing `Date & Time` field
image
*  Added Thermo Fisher Visionlite headers length validation
* Added Thermo Fisher Genesys on Board headers length validation
* Added Agilent Tapestation Analysis invalid xml validation
* Set Unchained Labs Lunatic excel parsing engine


src/allotropy/parsers/novabio_flex2/novabio_flex2_parser.py
```diff
@@ -27,9 +28,13 @@ class NovaBioFlexParser(VendorParser[Data, Model]):
...
         # NOTE: calling parse_dates and removing NaN clears empty rows.
-        data = read_csv(
-            named_file_contents.contents, parse_dates=["Date & Time"]
-        ).replace(np.nan, None)
+        try:
+            data = read_csv(
+                named_file_contents.contents, parse_dates=["Date & Time"]
+            ).replace(np.nan, None)
+        except Exception as e:
+            msg = "NovaBio Flex2 parser requires 'Date & Time' column to be present in the CSV file."
```
nathan-stender, line 36: We can't catch all exceptions from this call and return this error message.

If the call fails for another reason, the message will be misleading.

Instead, we should check for the "Date & Time" column first, and throw an error if missing.

src/allotropy/parsers/thermo_fisher_visionlite/thermo_fisher_visionlite_reader.py
```diff
@@ -31,9 +31,11 @@ def __init__(self, named_file_contents: NamedFileContents) -> None:
         if all(
             col_name not in first_line.lower() for col_name in expected_required_columns
         ):
-            self.header = SeriesData(
-                pd.Series(first_line.split(",")[:4], index=HEADER_COLS)
-            )
+            columns = first_line.split(",")[:4]
+            if len(columns) != len(HEADER_COLS):
+                msg = f"Expected {len(HEADER_COLS)} columns, but got {len(columns)}."
```
nathan-stender, line 36:
```suggestion
                msg = f"Expected {len(HEADER_COLS)} columns, but got {len(columns)}: {columns}."
```

nathan-stender (changes requested):

src/allotropy/parsers/novabio_flex2/novabio_flex2_parser.py
nathan-stender: Ah right, we don't have the pd columns yet.

What you could do is:

1. Only except ValueError
2. Check that the error message includes "Missing column provided to 'parse_dates'"

nathan-stender (changes requested):
Responded to question - requesting changes to remove my review request until addressed

nathan-stender (approved):

PR #955, opened 2025-04-08
feat: Cytiva Biacore T200 Control - fix missing stream data errors

slopez-b:
There were some errors in Cytiva Biacore Control files because of unhandled cases of missing data streams that are optionals, such as:
 - Report point data for the first cycle of manual runs, this made the adapter raise since it originaly expected all cycles to contain Report point data and Sensorgram data.
 - The `ApplicationTemplate` stream is not included in all experiments, this stream contains mainly optional metadata.


james-leinas (dismissed):
ASM LGTM 👍

nathan-stender (changes requested):
Code LGTM - quick nit

nathan-stender (approved):

PR #957, opened 2025-04-10
refactor: FlowJo - Add strategy pattern for getting region vertices

joshua-benchling:


src/allotropy/parsers/flowjo/flowjo_structure.py
```diff
@@ -110,6 +110,150 @@ class VertexRole(Enum):
     EDGE = "edge"
 
 
+class VertexStrategy:
```
nathan-stender, line 113: Nice abstraction!

A couple things:

1. Let's re-name to `VertexExtractor` to better describe what the class does.
2. Move the mapping logic to a build function in the base class, that will keep the logic encapsulated.

```
    @staticmethod
    def build(region_type: RegionType, gate_element: StrictXmlElement, x_dim: str | None, y_dim: str | None) -> VertexStrategy:
        ... move mapping stuff here ...
        return strategy_class(gate_element, x_dim, y_dim)
```

3. Do we have any more vertex types we know we are ignoring? If not, prob best to throw an AllotropyConversionError saying "gate type ... is not currently supported."

src/allotropy/parsers/flowjo/flowjo_structure.py
```diff
@@ -387,114 +531,23 @@ def _extract_vertices(
...
-        edge_vertices = gate_element.findall("gating:edge/gating:vertex")
-        for vertex in edge_vertices:
-            _add_vertex_with_role(vertex, VertexRole.EDGE.value)
+    strategy_class = strategies.get(gate_type)
+    if strategy_class:
+        strategy = strategy_class(gate_element, x_dim, y_dim)
+        return strategy.extract()
 
-    return vertices if vertices else None
+    return None
```
nathan-stender, line 550: After adding mapping logic to build function, this should become `return VertexStrategy.build(gate_type, gate_element, x_dim, y_dim).extract()` :D

nathan-stender (changes requested):

src/allotropy/parsers/flowjo/flowjo_structure.py
```diff
@@ -110,6 +110,173 @@ class VertexRole(Enum):
...
+            if y_dim is not None and y_dim.lower() == "time"
+            else TQuantityValueRelativeFluorescenceUnit.unit
+        )
+
+    def extract(self) -> list[Vertex] | None:
+        """Extract vertices from the gate element."""
+        raise NotImplementedError
+
+    @staticmethod
+    def build(
```
nathan-stender, line 138: build should return the `VertextExtractor` 

Then you can call `VertextExtractor.build(...args...).extract()`

nathan-stender (approved):

PR #960, opened 2025-04-10
feat: Molecular Devices SoftMax Pro -  remove enforcement of WellPlateName in group data for single plate experiments

slopez-b:
Currently the SMP adapter only reports calculated data documents when group blocks include the WellPlateName column.

Scenario 1:

Only 1 plate within export file

Remove requirement for WellPlateName to be present in group data blocks for recording calculated data – assumption being that references to wells will be for the plate within the export.

Scenario 2:

Multiple plates within export file

Keep enforcement for WellPlateName to be present in group data blocks to distinguish data source of wells within the experiment.


james-leinas (approved):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -1204,6 +1200,13 @@ def create(reader: CsvReader) -> StructureData:
         for group_block in block_list.group_blocks:
             for group_sample_data in group_block.group_data.sample_data:
                 for group_data_element in group_sample_data.data_elements:
+                    # For experiments with only one plate, it is assumed that all group blocks belong to it.
+                    if len(block_list.plate_blocks) == 1:
+                        group_data_element.plate = next(iter(block_list.plate_blocks))
+                    elif group_data_element.plate is None:
```
nathan-stender, line 1206: From the description - it seems that we should be throwing an error in this case, right?

nathan-stender (changes requested):

tests/parsers/moldev_softmax_pro/testdata/MD_SMP_absorbance_endpoint_partial_plate_example03.json
```diff
@@ -4796,7 +4796,7 @@
                             },
                             "measurement identifier": "MOLDEV_SOFTMAX_PRO_TEST_ID_40",
                             "sample document": {
-                                "sample identifier": "Quantitation D5",
+                                "sample identifier": "MPCTDN186",
```
nathan-stender, line 4799: Why does sample identifier change here?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
slopez-b: The idea of trying to match the group data with the block data is to create calculated data and have a reference for the data sources. If we can't identify the plate a group is associated with, we just don't use the group data to create calculated documents, but we still create measurements for the samples in the plate block. This change just try to support the case when we have single plates, since it is assumed than in that case all group data belongs to it.

cc @james-leinas

tests/parsers/moldev_softmax_pro/testdata/MD_SMP_absorbance_endpoint_partial_plate_example03.json
slopez-b: The default `sample identifier` uses the `well plate identifier` and the `location identifier` unless we have group data associated with the plate, in which case we use the `sample` column for that group.

nathan-stender (approved):

PR #982, opened 2025-05-06
fix: Beckman Echo Plate Reformat -  remove accidental volume conversion of survey volume results

ersnow42:
Survey volume results (for 'Survey Fluid Volume' and 'Current Fluid Volume' fields in the Echo output CSV) are already in microliters, and we were spuriously converting this number "from nl to ul" resulting in incorrect volumes in the output data.  
Also added another test case that has exceptions with valid survey results.

### DIFF
--- a/tests/parsers/beckman_echo_plate_reformat/testdata/ExceptionsOnly.json
+++ b/tests/parsers/beckman_echo_plate_reformat/testdata/ExceptionsOnly.json
@@ -45,7 +45,15 @@
                                 ]
                             },
                             "measurement identifier": "BECKMAN_ECHO_PLATE_REFORMAT_TEST_ID_0",
-                            "measurement time": "2024-12-09T13:29:31.401000+00:00"
+                            "measurement time": "2024-12-09T13:29:31.401000+00:00",
+                            "aspiration volume": {
+                                "value": 0.0,
+                                "unit": "μL"
+                            },
+                            "transfer volume": {
+                                "value": 0.0,
+                                "unit": "μL"
+                            }
                         },
                         {
                             "device control aggregate document": {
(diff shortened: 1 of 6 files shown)


james-leinas (dismissed):

tests/parsers/beckman_echo_plate_reformat/testdata/ExceptionsOnly.json
```diff
@@ -15,6 +15,8 @@
                                             "unit": "μL"
                                         },
                                         "custom information document": {
+                                            "survey fluid volume": 0.0,
+                                            "current fluid volume": 0.0,
```
slopez-b, line 19: @ersnow42 Changes look good to me. 

I have just a question about the original conversion. In the `convert_echo_nl_to_ul` function, the value is only returned if it is Truty:

```python
def convert_echo_nl_to_ul(value: float | None) -> float | None:
    return (
        (value * constants.PLATE_REFORMAT_REPORT_VOLUME_CONVERSION_TO_UL)
        if value
        else None
    )
```
This means that a value of `0.0` is removed in the conversion because it would fail the condition. My question is if there's a reason for this condition, or if it was an error in the original implementation. 

For consistency, we should either report only non zero values, or report all non Null (`None`) values in both cases (converted and non converted results).

cc @james-leinas

slopez-b (dismissed):

tests/parsers/beckman_echo_plate_reformat/testdata/ExceptionsOnly.json
nathan-stender: > In general in Allotropy parser world, do we prefer to omit the key entirely if there's a missing value in the input file, or return a None value for that key in the results dict?

We generally omit the key entirely if there's a missing value in the input file

The only exception is usually if we have a clear indication of what the error is, in which case we will put -0 (constant NEGATIVE_ZERO) and include an error document.

ersnow42:
OK, so that sounds like we need to distinguish between None and 0 so that
we actually report zeroes post conversion.

nathan-stender:
Yup agreed!

nathan-stender (approved):

PR #984, opened 2025-05-07
feat: Unchained Labs Lunatic - add unread data

james-leinas:


joshua-benchling (dismissed):

src/allotropy/allotrope/schema_mappers/adm/plate_reader/rec/_2024/_06/plate_reader.py
```diff
@@ -119,6 +119,7 @@ class Measurement:
     detection_type: str | None = None
     sample_role_type: SampleRoleType | None = None
     mass_concentration: float | None = None
+    custom_info: dict[str, Any] | None = None
```
nathan-stender, line 122: You added this, but are not reading from it.

It looks like the measurement already had a custom info field called "measurement_custom_info"

I think you should be adding to that in the unchained labs parser and remove this field.

nathan-stender (changes requested):

src/allotropy/parsers/unchained_labs_lunatic/unchained_labs_lunatic_structure.py
```diff
@@ -179,6 +211,36 @@ def create_metadata(header: SeriesData, file_path: str) -> Metadata:
...
+                "application",
+                "concentration (ng/ul)",
+                "background wvl. (nm)",
+                "plate id",
+                "plate position",
+            },
+            skip_regex={
+                # Strings to skip since these are already captured as measurements/calculated data
+                # Skip absorbance spectrum measurements with a### (10mm)
+                r"^a\d{3} \(10mm\)$",
```
joshua-benchling, line 233: Actually the skip list is being treated as a regex already, did you try combining and check if you get the same result?

james-leinas: You're right... no need for the addition. Just checked and tests passed with the combination

nathan-stender (approved):

PR #1000, opened 2025-05-22
fix: Molecular Devices SoftMax Pro - handle case where group data is missing value for all rows

nathan-stender:


src/allotropy/allotrope/converter.py
```diff
@@ -101,6 +101,7 @@
     "[": "_OBRACKET_",
     "]": "_CBRACKET_",
     "~": "_TILDE_",
+    "?": "_QMARK_",
```
nathan-stender, line 104: Note: this handles the question mark in the new custom information field introduced in the test file

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -164,7 +164,7 @@ def create(cls, data: pd.DataFrame, calc_data_cols: list[str]) -> GroupSampleDat
                             ),
                         )
                     )
-                else:
+                elif data_elements[column]:
```
nathan-stender, line 167: Previous behavior - if a row was missing a value for the given group data column, copy from the previous row.

This fails if the first row is empty. Handle this by checking if the existing list is empty.

stephenworlow (approved):

PR #1005, opened 2025-06-04
fix: BMG Mars: wrap keyError when index out of range

joshua-benchling:


src/allotropy/parsers/bmg_mars/bmg_mars_reader.py
```diff
@@ -27,7 +28,13 @@ def __init__(self, named_file_contents: NamedFileContents) -> None:
             axis="index"
         )
         new = df["value"].str.split(": ", expand=True, n=1)
-        self.header = SeriesData(pd.Series(new[1].values, index=new[0].str.upper()))
+
+        # Handle the case where no ": " delimiter is found, resulting in a DataFrame with only one column
+        if new.shape[1] < 2:
+            msg = "Unable to parse header data: no key-value pairs found with expected format."
+            raise AllotropeConversionError(msg)
+        else:
```
nathan-stender, line 36: Nitty I know, but no need for else here since you'll raise in if

nathan-stender (changes requested):

nathan-stender (approved):

PR #1013, opened 2025-06-11
fix: Molecular Devices SoftMax Pro - account for partial plate reads

joshua-benchling:


james-leinas (dismissed):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_data_creator.py
```diff
@@ -80,9 +84,12 @@ def _get_data_cube(
 
 def _get_spectrum_data_cube(
     plate_block: PlateBlock, data_elements: list[DataElement]
-) -> DataCube:
+) -> DataCube | None:
     wavelengths = [data_element.wavelength for data_element in data_elements]
     values = [data_element.value for data_element in data_elements]
+    if NEGATIVE_ZERO in values:
```
nathan-stender, line 90: This is causing us to drop any datacube with a `0` in it.

In `partial_plate_with_empty_values` we should be recording 360 wells, but we're only recording , we should be recording 360 wells, but we're only recording 352.

That's because `-0 == 0` evaluated to `True` in python.

I think we'll need to allow `value` to stay `None` in `DataElement` and then convert to `NEGATIVE_ZERO` after this check.

joshua-benchling: You are right, thank you.

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_data_creator.py
```diff
@@ -74,15 +78,23 @@ def _get_data_cube(
...
 
 
 def _get_spectrum_data_cube(
     plate_block: PlateBlock, data_elements: list[DataElement]
-) -> DataCube:
+) -> DataCube | None:
     wavelengths = [data_element.wavelength for data_element in data_elements]
     values = [data_element.value for data_element in data_elements]
+    if None in values:
+        # Ignore the wells completely from the ASM if there is an nan value
```
nathan-stender, line 96: Talked with @james-leinas internally about this.

We also don't want to drop a well entirely if there is a single NaN value, rather, we should make report the wavelengths with NaN values as error documents and use NEGATIVE_ZERO value, while reporting the rest of the data for that well.

I think your logic below that makes values NEGATIVE_ZERO already covers the creation of the data cube, so we just need to add error documents and change this clause to only return None if ALL values are None.

I don't think we actually have an example where a single value is missing, so we should modify (each well is currently all values or no values) so we should modify one of the new test files to exercise this case.

nathan-stender (changes requested):

PR #1016, opened 2025-06-17
feat: Molecular Devices SoftMax Pro - Enable Reduced-only file parsing

janedance:
Updating the Softmax parser to pull in the following data:

    Group: Control
    Sample	Wells	Sample#	Values	Results	MeanResults	CVs	CVresults	
    
    Group: Matrix_Spike_1
    Sample	Wells	Value	R	Result	MeanResult	CVs	CvResult	%SpikeRecovery	
    
    Group: Matrix_Spike_2
    Sample	Well	Value	R	Result	MeanResult	CVs	CVResult	%SpikeRecovery	
    
    Group: Standards
    Sample	Concentration ng/mL	Wells	Value	MeanValue	SD	CVs	CV_Result
    
    Group: Samples
    Sample	Wells	Value	R	Result	MeanResult	SD	CVs	CVresult

I believe this requires adding support for `REDUCED` files, so that's what this PR is for.

I think we also need the "Acceptance Criteria" data from the `Note' section, in which case I'll be following up with a larger PR to add some functionality there.


tests/parsers/moldev_softmax_pro/testdata/HCP_for_GB_JB.json
janedance, line 1: **blocking/suggestion**

Note to self: rename this file to something more meaningful.

janedance:
blocking/suggestion

Note to self: There is probably some documentation that needs to be updated along with this change.

nathan-stender (changes requested):
Hi Jane! Thank you so much for contributing!

Looking at this change, I think what we'll want to do for capturing only reduced data is a little different. It looks like this change is causing the reduced data to be captured as the raw data.

This causes us to report the raw data as the reduced data (you can see the reduced values are being reported as the "absorbance" values in the measurement document, whereas we want to report the reduced values as calculated data, since it's not really raw data!) 

I think to do this, what we actually want to do is check the read type in `TimeData` and only read the raw data if they read type is "raw" or "both", and only read the reduced data if the type is "reduced" or "both".

There is one more wrinkle specific to the ASM format, in that the measurement document currently requires an "absorbance" value, and we don't have that value for the reduced only format. 

I think the way to address this will be to, for the reduced only case, generate a`TimeRawData` with `TimeMeasurementData` with `NEGATIVE_ZERO` for the value and an ErrorDocument saying "Not reported"

Let me know if you have any questions!

janedance:
Thanks @nathan-stender for the detailed feedback, I appreciate it!

I made the requested change e86c186878b8f61a8e5624faef2e25ac64df3f75 and found that it had a wider impact than I anticipated. I *believe* this is because reduced data was previously being read for "RAW" type. So, the test output changed for several cases.

I want to be cautious here, since I didn't set out to change existing functionality. I would like to confirm with you:

- this change to existing functionality is what I think, and is a good thing
- stopping the reduced data being read for "RAW" type would not impact what would be available to the user (in other words, that this won't break any existing usage)

If it does have the possibility of breaking existing usage, then I can change it back to continuing reading reduced data for "RAW" type. Let me know what you think!

nathan-stender (commented):
You're right! I took a look at this, and it appears that the problem is that the RAW format can just... report reduced data as well 🤦. Confirmed this with our scientific consultant reviewing the existing files.

Given this, I think that we should keep the first part of what you added (reduced only creating empty data) but go back to always trying to read reduced data.

Given that, I think the existing cases will work as before, and your case will work!

Other than that, one minor request to extract the empty data generation to a util function and I think we'll be all set!

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -712,16 +712,19 @@ def create(
         wavelength: float,
         columns: pd.Series[str],
     ) -> TimeWavelengthData:
-        data = assert_not_none(
-            reader.pop_csv_block_as_df(sep="\t"),
-            msg="unable to find raw data from time block.",
-        )
-        set_columns(data, columns)
+        # Try to get the raw data block
+        data_df = reader.pop_csv_block_as_df(sep="\t")
```
nathan-stender, line 716: Gut check - I don't think we need these changes anymore given the new solution?

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -789,13 +792,69 @@ def create(
...
     ) -> TimeData:
+        raw_data = None
+        reduced_data = None
+
+        # Read raw data if data_type is RAW or BOTH
+        if header.data_type in (DataType.RAW.value, DataType.BOTH.value):
+            raw_data = TimeRawData.create(reader, header)
+        # For REDUCED only, create synthetic raw data with error message
+        else:
+            # Create synthetic wavelength data for each wavelength
```
nathan-stender, line 803: Code LGTM - request to pull this out into a helper just for clarity

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -789,13 +792,69 @@ def create(
...
+                        wavelength=wavelength,
+                        measurement_data=measurement_data,
+                    )
+                )
+
+            # Create TimeRawData with our synthetic wavelength data
+            raw_data = TimeRawData(wavelength_data=synthetic_wavelength_data)
+
+        # Read reduced data if data_type is REDUCED or BOTH
+        if (
```
nathan-stender, line 849: At noted in other reply, it appears that RAW can still have REDUCED data... so we can go back to always trying to read in REDUCED data if `reader.current_line_exists()`

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -874,8 +877,12 @@ def get_read_mode(cls, read_type: str, read_mode_raw: str) -> str:
...
     def check_data_type(cls, data_type: str) -> None:
-        if data_type not in (DataType.RAW.value, DataType.BOTH.value):
-            msg = f"The SoftMax Pro file is required to include either 'Raw' or 'Both' (Raw and Reduced) data for all plates, got {data_type}."
+        if data_type not in (
+            DataType.RAW.value,
+            DataType.BOTH.value,
+            DataType.REDUCED.value,
+        ):
+            msg = f"The SoftMax Pro file is required to include either 'Raw', 'Reduced', or 'Both' (Raw and Reduced) data for all plates, got {data_type}."
             raise AllotropeConversionError(msg)
```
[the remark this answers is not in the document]
nathan-stender: I think it makes sense to keep the check - just to throw a readable error if there's some value we don't expect (some new read type gets added)

That said, I think we can change the msg to something like:

`Unexpected data type: {data_type}, supported values are RAW, REDUCED, or BOTH`

nathan-stender (changes requested):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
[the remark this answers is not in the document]
janedance: Sounds good, updated in commit bed36c58ea430e93db082c4ef7c753a399a179cb

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
janedance: Agreed fc59b053cc41da4e17fceb3eb8f4f0d96ba8057a

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
janedance: Agreed fc59b053cc41da4e17fceb3eb8f4f0d96ba8057a

janedance:
Thanks @nathan-stender once more for the detailed feedback! 🙏 

I've made the requested changes - let me know if you have any other suggestions

nathan-stender (approved):

janedance:
> @janedance this looks perfect! Thank you so much for the contribution!
> 
> One last thing - feel free to add yourself to the `authors` list in `pyproject.yaml` if you want to!

You're very welcome! And thank you for the code reviews!

I've added myself to the authors list :)

PR #1027, opened 2025-07-02
feat: Molecular Devices SoftMax Pro - move maximum wavelength signal to measurement document

slopez-b:
Reporting `maximum wavelength signal` within the `plate reader document` was causing issues when parsing ASM in the asm-to-csv converter since the metadata at the "technique document" level is assumed to be the same for all documents.

This change also makes sense since it is possible to have several reads per well in `spectrum` plates (not supported yet), in which case a different `maximum wavelength signal` would be reported per `measurement document`


nathan-stender (approved):

nathan-stender:
Closing per external discussion - we will be moving max wavelength to calc docs

PR #1035, opened 2025-07-15
feat: Luminex Intelliflex - Create connector

joshua-benchling:


nathan-stender (changes requested):
It looks like the reader/data implementation can be kept as is from the other parser.

We should try to reduce the code as much as possible (can always add overrides later if we discover diffs)

Here, it looks like we really just need a separate `create_metadata` function, and can reuse the rest as-is

nathan-stender (approved):

PR #1076, opened 2025-09-16
feat: Migrate ctl_immunospot to use SeriesData.get_unread

oscar-castro-bl:


james-leinas (changes requested):
### CTL ImmunoSpot UnRead Data

_Skip_

- Authenticated User -- already mapped to analyst

_Measurement Aggregate Document_

- Assay
- Counted
- Review Date

_Device Control Document_

- Edge Effect Compensation
- Edge Compensation Level
- Normalize Counts of Max
- Objects
- Diffuseness
- Counting window size
- Counting mask size (%)
- Background Balance
- Auto Areas
- Manual Areas
- Scanning
- Counting

_Image Feature Document_

- Min. SpotSize
- Max. SpotSize
- Spot Separation

james-leinas (approved):

oscar-castro-bl:
@nathan-stender would you mind giving it a look? thanks

tests/parsers/ctl_immunospot/testdata/ctl_immunospot_v7_0_38_8_QC.json
```diff
@@ -28,6 +33,7 @@
                             "processed data aggregate document": {
                                 "processed data document": [
                                     {
+                                        "data processing document": {},
```
james-leinas, line 36: Just caught this, can you ignore adding the data processing document if no values found in the dict?

src/allotropy/parsers/ctl_immunospot/ctl_immunospot_reader.py
```diff
@@ -213,13 +211,45 @@ def fix_line(line: str) -> str:
                 line = f"Review Date: {date_str}"
             return line.strip()
 
-        lines = [fix_line(line) for raw_line in lines for line in raw_line.split(";")]
+        def split_multi_key_line(line: str) -> list[str]:
+            """Split lines that contain multiple key-value pairs into separate lines."""
+            line = line.strip().strip('"').strip()  # Clean up quotes and whitespace
```
nathan-stender, line 216:
```suggestion
            line = line.strip(' "')  # Clean up quotes and whitespace
```

src/allotropy/parsers/utils/pandas.py
```diff
@@ -279,7 +279,11 @@ def _key_matches(self, match_key: str, key: str) -> bool:
         # "++" can cause re.compile to fail. Since it is never a valid regex expression
         # it is safe to escape it to prevent the error.
         match_key = match_key.replace("++", r"\+\+")
-        return bool(re.fullmatch(match_key, key))
+        try:
+            return bool(re.fullmatch(match_key, key))
+        except re.error:
```
nathan-stender, line 284: I'm concerned about silently skipping failing regexes, as this would mean a bad regex added to a parser would silently fail and just not pick up data.

I would prefer to diagnose the failing regex and fix it, as we do for ++ above.

That said, I checked out your branch and removed this to check the error myself, and found that there were no failures. 

I'm guessing that your other updates made it so we're no longer trying the bad regex? So I think OK to remove this for now.

nathan-stender (changes requested):
Code looks good! A couple small changes.

Also looks like James has a request

src/allotropy/allotrope/converter.py
```diff
@@ -150,6 +152,9 @@ def add_custom_information_document(
             continue
         cleaned_dict[key] = value
 
+    # Filter out Python keywords - they can't be used as dataclass field names
```
nathan-stender, line 155: We don't want to filter out keys silently. We have a mechanism for handling invalid model keys with 
`_convert_model_key_to_dict_key` and `_convert_dict_to_model_key`

I think you could handle this like this:

```
def _convert_model_key_to_dict_key(key: str) -> str:
    key = SPECIAL_KEYS.get(key, key)
    if key.startswith("_KW"):
        key = key[3:]
    if key.startswith("___") and key[3].isdigit():
        key = key[3:]
    for dict_val, model_val in DICT_KEY_TO_MODEL_KEY_REPLACEMENTS.items():
        key = key.replace(model_val, dict_val)
    return key


def _convert_dict_to_model_key(key: str) -> str:
    key = SPECIAL_KEYS_INVERSE.get(key, key)
    if keyword.iskeyword(key):
        key = f"_KW{key}"
    if key[0].isdigit():
        key = f"___{key}"
    for dict_val, model_val in DICT_KEY_TO_MODEL_KEY_REPLACEMENTS.items():
        key = key.replace(dict_val, model_val)
    return key
```

nathan-stender (changes requested):

PR #1081, opened 2025-09-19
feat: Migrate beckman_pharmspec to use SeriesData.get_unread

oscar-castro-bl:


ajcariaga16 (approved):

src/allotropy/parsers/utils/pandas.py
```diff
@@ -269,6 +269,10 @@ def __del__(self) -> None:
                 )
 
     def _get_custom_key(self, key: str) -> float | str | None:
+        # Skip non-string keys (like nan, None, etc.) as they can't be used as custom info keys
```
nathan-stender, line 272: Please remove - want to avoid silent skipping of values in this library.

A None or invalid key is usually a problem with the parser itself.

In this case, it looks like you already addressed whatever was causing this in the parser, because removing this does not cause any test failures.

nathan-stender (changes requested):
Mostly LGTM - one change

PR #1095, opened 2025-10-02
feat: Cytiva Biacore T200 - Refactor to use JsonData

felipenarv:
Basically replaced DictType with JsonData, which keeps track of which keys have been called and create a warning when the object is deleted without having read all keys

### DIFF
--- a/src/allotropy/parsers/cytiva_biacore_t200_control/cytiva_biacore_t200_control_structure.py
+++ b/src/allotropy/parsers/cytiva_biacore_t200_control/cytiva_biacore_t200_control_structure.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from dataclasses import dataclass, field
+from typing import Any
 
 import pandas as pd
 
@@ -13,6 +14,7 @@ from allotropy.allotrope.schema_mappers.adm.binding_affinity_analyzer.benchling.
 )
 from allotropy.parsers.constants import NOT_APPLICABLE
 from allotropy.parsers.cytiva_biacore_t200_control import constants
+from allotropy.parsers.utils.json import JsonData
 from allotropy.parsers.utils.pandas import map_rows, SeriesData
 from allotropy.parsers.utils.uuids import random_uuid_str
 from allotropy.parsers.utils.values import (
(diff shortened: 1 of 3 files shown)


src/allotropy/parsers/utils/dict_data.py
```diff
@@ -0,0 +1,304 @@
...
+            msg = f"Expected dict, got {type(data)}"
+            raise AllotropeConversionError(msg)
+        super().__init__()
+        self._read_keys: set[str] = set()
+        self.errored = False
+        self.creation_stack = traceback.extract_stack()
+        for key, value in data.items():
+            super().__setitem__(key, self._wrap(value))
+
+    def _wrap(self, value: Any) -> Any:
```
joshua-benchling: Since this is doing the same as `from_any`, we could do:
```python3
def _wrap(self, value: Any) -> Any:
    return DictData.from_any(value)
```

src/allotropy/parsers/cytiva_biacore_t200_control/cytiva_biacore_t200_control_structure.py
```diff
@@ -81,35 +82,56 @@ class RunMetadata:
...
-            ),
-            baseline_flow=try_float_or_none(
-                application_template_details.get("BaselineFlow", {}).get("value")
-            ),
-            data_collection_rate=try_float_or_none(
-                application_template_details.get("DataCollectionRate", {}).get("value")
+                else application_template_details.get(
+                    DictData,
+                    "system_preparations",
+                    DictData({}),
```
joshua-benchling: NIT but you can just pass ` DictData()` instead of ` DictData({})` . This is slightly cleaner.

joshua-benchling (dismissed):
Good implementation for wrapping up nested dicts. Left optional comments

nathan-stender (dismissed):

src/allotropy/parsers/cytiva_biacore_t200_control/cytiva_biacore_t200_control_structure.py
felipenarv: Done. I added this function and replaced it in the places it would work

src/allotropy/parsers/utils/dict_data.py
felipenarv: I removed "from_any", that was a function I used for another logic before setting it as it is, so it is not necessary or even useful

PR #1240, opened 2026-07-13
fix: Support SpectraMax 340PC-384 kinetic files in SoftMax Pro parser

nathan-stender:
## Summary
- SpectraMax 340PC-384 kinetic export files failed to parse due to multiple format differences from previously supported files
- **Encoding**: Parser now uses chardet detection instead of assuming UTF-8 (files are ISO-8859-1)
- **Empty groups**: Group blocks with no data rows (e.g. empty "Standards" groups) are handled gracefully
- **Partial plates**: Wells empty at t=0 are skipped in kinetic mode instead of erroring; wells that disappear at later timepoints still produce an error
- **Time format**: `time_to_seconds` now accepts `M:SS` and `H:MM` in addition to strict `HH:MM:SS`
- **GroupColumns**: Files without `Group Column`/`Formula Name` sections no longer error

## Test plan
- [x] All 36 existing moldev_softmax_pro tests pass
- [x] Customer file (`46_SpectraMax340PC-384.txt`) parses successfully
- [x] Missing kinetic measurement error test updated to validate the correct error path (well populated at t=0 but missing at later timepoint)
- [x] Lint clean (ruff, black, mypy)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
```diff
@@ -65,14 +64,25 @@ def num_wells_to_n_columns(well_count: int) -> int:
...
+    """Transforms time string into seconds. Supports H:MM:SS, HH:MM:SS, H:MM, M:SS formats."""
+    parts = time.split(":")
+    if len(parts) == 3:
+        try:
+            h, m, s = int(parts[0]), int(parts[1]), int(parts[2])
+            return h * 3600 + m * 60 + s
+        except ValueError as e:
+            msg = f"Bad time formatting, expected HH:MM:SS, got {time}"
+            raise AllotropeConversionError(msg) from e
+    elif len(parts) == 2:
```
ajcariaga16, line 76: Comment says supports `H:MM`, but from reading this, i don't think that's correct

ajcariaga16: Not sure how we'd differentiate between those though :p probably just a confused comment

ajcariaga16 (dismissed):

src/allotropy/parsers/moldev_softmax_pro/softmax_pro_structure.py
nathan-stender: Yeah nice catch, fixed

stephenworlow (approved):

PR #1245, opened 2026-07-23
fix: Support EnVision calculated formulas without a name prefix

nathan-stender:
## Summary
A PerkinElmer EnVision 384-well absorbance export (`OD450`) failed to parse:

```
AllotropeConversionError: Unable to find expected formula name for calculated results section.
```

The calculated plate's `Formula` is a plain description with no `name =` prefix:

```
Calc 1: OD450 direct absorbance value at 450 nm
```

`CalculatedPlateInfo.create` extracted the name with `^([^=]*)=`, which assumes the standard form `Calc N: <name> = <expression>` and raised when no `=` was present.

## Fix
When the formula has no `=`, fall back to using the whole formula string as the calculated data name. Standard formulas are unaffected.

## Changes
- `perkin_elmer_envision_structure.py`: make the formula-name regex optional with a whole-string fallback.
- Reworked `test_create_calculated_plate_info_with_invalid_formula` → `test_create_calculated_plate_info_with_formula_without_name` to assert the fallback.
- Added regression file `PE_Envision_absorbance_OD450_384_calc_no_name.csv` (+ expected JSON).

## Testing
- The file now parses (384 wells, 384 calculated data documents; calc name = full formula string).
- All 25 `perkin_elmer_envision` tests pass; `hatch run lint` (ruff, black, mypy) clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nathan-stender:
Superseded by #1246, which drops the large 384-well testdata file and instead exercises the no-'=' formula case by editing an existing example file.

PR #1267, opened 2026-09-16
fix: Luminex xPONENT - report NaN assay bead counts as the NaN sentinel

ajcariaga16:
### Problem

A NaN assay bead count in a Luminex xPONENT file reaches the ASM as a bare float `NaN`. It passes schema validation (jsonschema treats `nan` as a `number`) but downstream consumers render it as a **blank cell** rather than `"NaN"`, inconsistent with other NaN-valued columns. (Benchling currently post-processes this with a workaround; this fix would retire that.)

### Fix

- `luminex_xponent_structure.py`: build `assay_bead_count` via `try_float_or_nan(...)` at both the analyte and total-events sites, so a NaN surfaces as `InvalidJsonFloat.NaN` (serialized `"NaN"`).
- multi-analyte-profiling mapper: `assay_bead_count` field types `float` → `JsonFloat`.
- **Schema:** the model's `TQuantityValue.value` is already `JsonFloat`, but the schema restricts `tQuantityValue.value` to `{type: number}`, so the sentinel fails validation. This loosens it to `anyOf[number, enum["NaN","+Infinity","-Infinity"]]`, aligning the schema with the model.

### Testing

- New `luminex_xPONENT_nan_bead_count` case: schema validation passes and NaN bead counts serialize as `{"value": "NaN", "unit": "#"}`.
- Full Luminex xPONENT suite: **50 passed**, no changes to any existing expected output.

### ⚠️ Design decision for maintainers

1. **Global vs targeted.** I loosened the shared `tQuantityValue` (all quantity values may now be `NaN`) because it simply makes the schema match the already-`JsonFloat` model. The narrower alternative is a dedicated nullable quantity-value type applied only to `assay bead count`. Your call on the standard direction (this diverges from published Allotrope core `{type: number}`).
2. **Bundled-schema regen.** I edited `core.schema.json` (source) and hand-matched the embedded copy in the multi-analyte-profiling bundle to demonstrate it end-to-end. Finalizing should run `hatch run scripts:generate-schemas` so **all** bundled schemas are regenerated consistently from the core change (I did not run the full rebundle here).


nathan-stender (changes requested):
We can't change the REC ASM schemas, as that will cause the output to fail to validate against the ASM schema. 

Will follow up offline with ideas for alternatives.

ajcariaga16:
Superseded by #1268. Per Nate: we can't edit the REC `tQuantityValue` schema. The correct pattern (already used for NaN statistics) is NEGATIVE_ZERO + an error document, which the ASM→CSV layer renders as "NaN" — no schema change. #1268 implements that.

